PR #16, opened 2017-03-12
Added jquery and queue framework (styling and necessary js and php files)

JakeKo:


2mnyzs:
Can you move the queue.php code from assets/server into index.php?

I think that the assets/server is for callbacks/backend focused PHP code.

PR #18, opened 2017-03-13
additional test appointment

jcowman2:


jcowman2:
Matt, I addressed all your suggestions. Thanks for the help! Let me know if anything else needs fixing

PR #19, opened 2017-03-28
Client queue back end

MatthewMeacham:
Implemented the client queue back-end for loading the appointments from the database and displaying them on the screen


JakeKo:
LGTM Approved

PR #28, opened 2017-05-07
Appointment db rework

MatthewMeacham:
I may have merged that branch into this one before creating this PR

- Added a inputType column to Question to specify what sort of input type is should be displayed as on the client-side
- Added a placeholder column to Question (which will be the placeholder in a text field or the default selected in a select input)
- Added a tag column to Question (for ease with sql queries, instead of necessitating a question id or the exact text, we will instead query by this shorthand 'tag' column. For example, first_name, last_name, or military_status)
- Added a subheading column to Questions which can be used to group the questions together by the type of information they are trying to get
- Split up the "What is your name" question into two questions: "What is your first name" and "What is your last name"

PR #27, opened 2017-04-30
Loginout

2mnyzs:
Login Logout funcitonality
- MySQL script to create needed tables
- Page for logging in and requesting a password reset
- Page for resetting password

PR #29, opened 2017-05-07
01 working queue

MatthewMeacham:
DO NOT MERGE UNTIL YOU'VE MERGED #27 and #28 
Again, I merged those into this branch before proceeding ^_^

- Queue now loads immediately on page load
- Queue now pulls data from database and displays it on the page
- Added two additional test appointments so you can see this work in action. They are both for 5/7/2017 with the first one starting at 12:30pm, so you will have to change the date if testing after 7:30pm on 5/7/2017.


server/queue.php
```diff
@@ -3,20 +3,46 @@
...
+			FROM AppointmentQuestionAnswer AS AQA
+			JOIN Answer ON AQA.answerId = Answer.answerId
+			JOIN Question ON Answer.questionId = Question.questionId
+			WHERE Question.tag = "first_name" AND AQA.appointmentId = ?');
+
+		$lastNameStmt = $conn->prepare('SELECT Answer.string AS lastName
+			FROM AppointmentQuestionAnswer AS AQA
+			JOIN Answer on AQA.answerId = Answer.answerId
+			JOIN Question ON Answer.questionId = Question.questionId
+			WHERE Question.tag = "last_name" AND AQA.appointmentId = ?');
```
CollinsSpencer: Do we need to prepare a statement for every person in the queue? These can just be put before the foreach loop.

CollinsSpencer: With setting them null after the foreach block as well. Is it safer in some way to prepare the statement multiple times? I thought a prepared statement was to avoid that?

MatthewMeacham: I'm going to reoptimize after this was pushed, but yah you're 100% right. I'll pull those out for right now, but I plan on getting it all into one statement anyway.

MatthewMeacham: done

PR #30, opened 2017-05-07
Appointment signup

CollinsSpencer:
MAKE SURE YOU MERGE #27 #28 AND #29 FIRST.

I got the signup form working, loading from the database.  I also left my sample page in as sampleSignupForm.php. It can be deleted at a later point.

Things still needed:
 - If you look at the sample, some of the questions have hints.  I think those are important so we should probably add them to the database.
 - I only have textfields and selects implemented. I have things like checkboxes in the sample, so those would be easy to add later.
 - The form doesn't do anything yet, but it at least loads from the database.

Note:
 - The signup form is split between files named signup and files named form. This is so that the form aspect is more reusable for future projects. With that in mind, some things may still need to be changed, such as the ajax in both signup.js and form.js. It isn't pretty like that, but it works for now.


CollinsSpencer:
I'm not sure if we need tags.  The inputTypes and subheadings are used.  However, tags are not used.  

That being said, we may want to consider adding an validation type column. For form validation (which is not currently being done with the form loaded from the database) it uses the name attribute, which I have consistently been making match the css id attribute. Currently they are formatted as vita#.  With jQuery validation, things like email, phoneUS, stateUS, zipcodeUS, date, etc. can be uniquely validated, making a validation type field useful.

Another column to consider adding is an error message column. Each field can have a unique error message, or if one isn't provided, there is a default.

MatthewMeacham:
@CollinsSpencer Fix the conflict you have in signup/index.php please.

The tag column is not for use on the front-end, that is a back-end thing that will help with SQL queries. Because, we can't query a question by its id since that can change when we re-evolve the database, nor can we query a question by the text is contains since that is subject to change, nor can we query a question by a placeholder because those are potentially not unique and subject to change. To solve that, the tag column was created as a static column that would never change (ideally) and would be globally unique (I'll add a constraint). 

The other two suggested columns can easily be implemented. I'll submit in another PR.

server/form.php
```diff
@@ -0,0 +1,39 @@
...
+			WHERE questionId = ' . $questionId);
+		$optionsStatement->execute();
+		$results = $optionsStatement->fetchAll();
+
+		$options = [];
+		foreach ($results as $result) {
+			$options[] = $result;
+		}
+
+		// $options = array("1","2","3");
```
MatthewMeacham: Don't push commented out code like this

signup/sampleSignupForm.php
```diff
@@ -0,0 +1,169 @@
...
+          <span class="vita-form-hint">(2 character state code; e.g. NE for Nebraska)</span>
+        </div>
+
+        <div class="vita-form-textfield">
+          <input class="" type="text" name="zip" id="zip" required>
+          <span class="vita-form-bar"></span>
+          <label class="vita-form-label vita-form-required" for="zip">Zip / Postal Code</label>
+        </div>
+
+        <!--
```
MatthewMeacham: Again, no commented out code por favor

assets/js/form.js
```diff
@@ -0,0 +1,72 @@
...
+    $("." + containingClass + " form").append("</select>" +
+  	      "<div class='vita-form-select__arrow'></div>" +
+          "</div>");
+  } else {
+    // Optional Select
+    $("." + containingClass + " form").append("<div class='vita-form-select'>" +
+          "<label for='" + id + "' class='vita-form-label'>" + label + "</label>" +
+          "<select id='" + id + "'>");
+
+    addOptions(containingClass, id, questionId);
```
MatthewMeacham: Can this not be pulled out of the if/else statement? It does the same thing as the one above

assets/js/form.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function newSelect(containingClass, id, label, questionId, required) {
+  if (required == true) {
+    // Required Select
+    $("." + containingClass + " form").append("<div class='vita-form-select'>" +
+          "<label for='" + id + "' class='vita-form-label vita-form-required'>" + label + "</label>" +
+          "<select id='" + id + "' class='required'>");
+
+    addOptions(containingClass, id, questionId);
```
MatthewMeacham: Can this not be pulled out of the if/else statement? It does the same thing as the one below

assets/js/form.js
```diff
@@ -0,0 +1,72 @@
...
+function newFormTitle(containingClass, title) {
+  $("." + containingClass + " form").append("<h1 class='vita-form-title'>" + title + "</h1>");
+}
+
+function newSubheading(containingClass, subheading) {
+  $("." + containingClass + " form").append("<h2 class='vita-form-subheading'>" + subheading + "</h2>");
+}
+
+function newTextField(containingClass, id, label, required) {
+  if (required == true) {
```
MatthewMeacham: Is there any way that we could make this so there's not so much duplicate code? Because the only part that changes in this if/else statement is the class='vita-form-required'.. Could we not just do the part before this, and then do an if(required) { append that part }
finish it off?

If it hinders readability too much though, what you have now looks good

signup/index.php
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html class="no-js" lang="">
+  <head>
+  <title>Signup Test</title>
+  <link href="https://fonts.googleapis.com/css?family=Montserrat" rel="stylesheet">
```
JakeKo: I'll probably end up moving these font imports into a main css file... It's faster and helps separate skin from structure.

Issue #31, opened 2017-05-07
Redesign Question Table


MatthewMeacham:
Redesign the Question table to become two tables: one to hold the front-end type information, and the other to hold the back-end type information. Front-end will contain the information like placeholder, inputType, validationType, hint, errorMessage. And the back-end will contain the information like questionId, questionString, tag, required, archived.

PR #32, opened 2017-05-07
31 redesign question database

MatthewMeacham:
I redid a bunch of the database after noticing a few *substantial* problems:
- **Renamed AppointmentQuestionAnswer to Answer**
- There is now a QuestionInformation table which contains the front-end type information related to questions such as placeholder, inputType, errorMessage, etc. (See #31)
- Question table contains only stuff directly related to questions now. The tag column can be used to query a specific question rather than having to use an id or the exact phrase of the question.
  - This tag column is globally unique and should be under_camel_case.
- I reorganized the "mysql scripts" folder
  - The "table_creation" folder contains scripts for creating the tables
  - The "test_data" folder contains scripts for inserting test data (development only)
  - The "production_data" folder will contain scripts for populating the important tables in production (namely Question and PossibleAnswer)
- There was a huge problem with the Answer table. In @CollinsSpencer 's method for populating select input types in his form.php, it queried the Answer table where the id was the question id. However, we are also inserting user responses into this Answer table. So right there, there was a massive problem in that select fields would have multiple of the same choices (unless we used a DISTINCT in the SQL query). However, beyond that, it was just a bad setup in my opinion.
  - This problem was remedied by splitting Answers into a new table and utilizing the old AppointmentQuestionAnswer (now called Answer) table: PossibleAnswer and Answer. 
  - PossibleAnswer contains the choices that we want to put into a select input type
  - Answer contains the actual user responses as well as the question and the appointment for which the answer is for
      - I am open to change the names of those tables.
- Fixed the queue after this new database redesign

@CollinsSpencer Will need to fix his PR in #30 after this gets merged. It will mainly affect your form.php with the select options, and also it will affect how you display questions, you will have to JOIN QuestionInformation now in order to get the placeholder, inputType, etc.

I began documenting this new database structure here:
https://docs.google.com/document/d/1BLd-y7Qpx02O5YR5Xaq8Y_sZgVDL5lHeYd2Sns9RuXY/edit?usp=sharing

And here is the new ERD so you can see generally what is going on.
appointmentdberd


mysql scripts/test_data/appointmentDBTestData.sql
```diff
@@ -0,0 +1,537 @@
+USE vita;
+
+-- sample questions with choices, if applicable
+INSERT INTO QuestionAccessories (inputType, placeholder, subheading, validationType, hint, errorMessage)
+	VALUES ("text", "First Name", "Contact Information", NULL, NULL, NULL);
+    
+INSERT INTO Question (string, tag, required, questionAccessoriesId) 
+	VALUES ("First Name", "first_name", TRUE, 
+    (SELECT questionAccessoriesId FROM QuestionAccessories
```
CollinsSpencer: Inconsistent spaces and tabs

server/config.php
```diff
@@ -10,7 +10,7 @@
 		$db_user = 'root';
 		$db_pass = 'root';
 
-		$DB_CONN = new PDO('mysql:host=127.0.0.1;dbname=vita;', $db_user, $db_pass);
+		$DB_CONN = new PDO('mysql:host=127.0.0.1:8889;dbname=vita;', $db_user, $db_pass);
```
CollinsSpencer: The default mysql port is 3306, so that's what it would probably be for production. I vote leave it without the port number and just change your local server to use 3306. That way there will be less to change from development to production.

mysql scripts/table_creation/appointmentDB.sql
```diff
@@ -0,0 +1,74 @@
...
+
+CREATE TABLE QuestionAccessories (
+	questionAccessoriesId INTEGER PRIMARY KEY NOT NULL AUTO_INCREMENT,
+    inputType VARCHAR(255) NOT NULL,
+    placeholder VARCHAR(255) NOT NULL,
+    subheading VARCHAR(255) NOT NULL,
+    validationType VARCHAR(255),
+    hint VARCHAR(255),
+    errorMessage VARCHAR(255)
+);
```
CollinsSpencer: Inconsistent spaces and tabs

mysql scripts/test_data/appointmentDBTestData.sql
MatthewMeacham: To be fixed when issue #33 is done

mysql scripts/table_creation/appointmentDB.sql
MatthewMeacham: To be fixed when issue #33 is done

server/config.php
MatthewMeacham: That's my bad, this file wasn't supposed to be pushed. Will fix now

mysql scripts/table_creation/appointmentDB.sql
```diff
@@ -0,0 +1,74 @@
...
+
+DROP TABLE IF EXISTS AppointmentQuestionAnswer;
+DROP TABLE IF EXISTS Appointment;
+DROP TABLE IF EXISTS Location;
+DROP TABLE IF EXISTS UserAnswer;
+DROP TABLE IF EXISTS PossibleAnswer;
+DROP TABLE IF EXISTS Question;
+DROP TABLE IF EXISTS QuestionAccessories;
+
+CREATE TABLE QuestionAccessories (
```
JakeKo: What do you think about QuestionDesign, QuestionStructure, QuestionDisplay, QuestionProperties, or QuestionTraits? Any of those boil your soup? I'd like to obey plurality if at all possible so maybe not the last two.

JakeKo: Additionally, this table could include html attributes like spellcheck, required, readonly, disabled, etc. It may seem excessive but if we're going this far to engineer this form, we might want to consider adding more of these constraints.

The alternative I see is to outsource all of the validation and constraint to @CollinsSpencer jQuery validation voodoo.

MatthewMeacham: The field "required" is in Question table. I'll defer to @CollinsSpencer if the spellcheck, readonly, disabled, etc. fields are necessary -- I don't want to put in useless columns if there's no practical client-side use for them. Validation is outsourced to @CollinsSpencer JQuery stuff.

PR #41, opened 2017-05-12
Basic Appointment Signup Form Submission

CollinsSpencer:
The full form submission is not yet complete. This PR is more of a sample of the form submission to show how the data will be handled. Heavy-lifting can be dealt with later.

Adding content to the database will be easy.

However, we should probably first discuss back-end form validation.

A large portion of the form submission is the appointment reminder system, so that will be implemented in a future PR.


assets/js/signup.js
```diff
@@ -23,7 +11,7 @@ var loadQuestions = function() {
   });
 
   $.getJSON({
-    url: '/server/form.php?retrieve=questions',
+    url: '/server/form.php?retrieve=questions&subheadings[]=Contact Information&subheadings[]=Language Information&subheadings[]=Background Information',
```
MatthewMeacham: Why do you have to specify the subheading[] for this? I'd really like for this to not be hard-coded, but I wanna hear the reasoning for this.

assets/js/signup.js
```diff
@@ -90,13 +113,18 @@ var questionRules = function(question) {
     rule["required"] = true;
   }
   if (question.validationType != null) {
+    // There should really be some form of validation for the validationType
+    // One option would be to make a new table with all of the possible validation types
+    // which would be both a foreign key for the validationType field in the questionInformation table
+    // and a way to validate at this point too. If the validationType has a typo for any
+    // given field, the front-end form validation fails to run at all for any of the fields.
```
MatthewMeacham: Why does one typo cause all the rest to fail? Can't you just like wrap each one in a try-catch or something..?

I suppose I'm not opposed to adding a ValidationType table (or maybe it could just be an enum, that would solve your problem too), however, it does seem a little over-engineered.. How many of the fields are not unique validations? Like phoneUS will probably only be used once... something like email would only be used once.. I don't see many instances where there would be repeat validationTypes on Questions, so I wouldn't want to add something like this. Although, I may consider the enum option, I just don't know how nicely this would play with evolving the database when we add a new validationType..

server/form.php
```diff
@@ -27,7 +37,7 @@
 	} else if ($retrieve == "options") {
 		$questionId = $_REQUEST['questionId'];
 
-		$optionsStatement = $conn->prepare('SELECT answerId, string, archived FROM Answer
+		$optionsStatement = $conn->prepare('SELECT possibleAnswerId, string, archived FROM PossibleAnswer
 			WHERE questionId = ' . $questionId);
```
MatthewMeacham: Uh isn't this bad... You're using a PreparedStatement but then just appending the variable part. You really should use the execute(array(variables)) method I think, as to prevent SQL injection

assets/js/signup.js
CollinsSpencer: Yeah. We don't currently have a method of imposing an ordering on the subheadings. That might be something that we would need a new table for. I did this for now so that the specific subheadings would not (even though they are currently hard-coded) be sitting in the form.php file. The ordering before now was alphabetic by subheading, but that isn't a very practical or useful solution for this.

MatthewMeacham: Understood, I could add an index to the questions and also a subheading index which specifies the ordering of questions and of subheadings, respectively.

assets/js/signup.js
CollinsSpencer: The validation methods include...

required, remote, email, url, date, dateISO, number, digits, equalTo, maxlength, minlength, rangelength, range, max, min, step
as found in https://github.com/jquery-validation/jquery-validation/blob/master/src/core.js

as well as 50 others found in https://github.com/jquery-validation/jquery-validation/tree/master/src/additional

That's 66 different possible validation types.

We would just need to check if the validationType we get back from the database is one of those strings, and then add it to the rule array if it is or don't if it isn't.

MatthewMeacham: I see. How many of these would actually be used though? And how many would be used more than once? Because, if we want functionality for all of those, then we will have to make this validationType thing support multiple for a single question (as is the case with min, max, and step -- all of those would be on a number input). 

Adding a ValidationType table (or enum) seems over-engineered to me since more likely than not, we would only use each validation type once.

server/form.php
```diff
@@ -5,20 +5,30 @@
 	$retrieve = $_GET['retrieve'];
 
 	if ($retrieve == "questions") {
-		// TODO make this handle multiple locations, if necessary
-		$questionStatement = $conn->prepare('SELECT questionId, string, inputType, placeholder, tag, subheading, required, archived FROM Question
-			WHERE (archived != true)
+		$subheadings = [];
```
JakeKo: This block of subheadings code seems awfully convoluted for what amounts to filling an array. Could we not try to guarantee that `$_GET['subheadings']` will always return an array.

More conditionals means slower page load. While this isn't a significant time suck, if it's easy to fix, fix it.

server/form.php
```diff
@@ -5,20 +5,30 @@
...
-			$result["errorMessage"] = "Sample error message.";
-			$result["validationType"] = "email";
-			$result["hint"] = "(Sample Hint)";
+		foreach ($subheadings as $subheading) {
+			$questionStatement->execute(array($subheading));
+
+			$results = $questionStatement->fetchAll();
 
-			$questions[] = $result;
+			foreach ($results as $result) {
```
JakeKo: Consider:
```php
$results = $questionStatement->fetchAll();
$questions = array_merge($questions, $results);
```

Alternatively:
```php
$questions = array_merge($questions, $questionStatement->fetchAll());
```


[Here](http://php.net/manual/en/function.array-merge.php) is the documentation for `array_merge()`.
Just some quality of life code cleanup :P

assets/js/signup.js
JakeKo: Might this be a config file solution?

assets/js/signup.js
```diff
@@ -79,6 +68,40 @@ var loadQuestions = function() {
       validationObject["rules"] = rules;
       validationObject["messages"] = messages;
       validateForm(validationObject);
+
+      //Since non-required fields are "valid" when they are empty, we need an
+      //alternate way to keep labels raised when there is content in their
+      //associated input field
+      $( ".vita-form-textfield input" ).focusout(function() {
+        if($.trim($(this).val()).length > 0) {
```
JakeKo: Consider:
```js
var isBlank = $.trim($(this).val()).length > 0;
$label = $(this).closest(".vita-form-textfield").find(".vita-form-label").toggleClass( "vita-form-label__floating", isBlank );
```
jQuery's `toggleClass()` has a second parameter option. Documentation can be found [here](http://api.jquery.com/toggleclass/).

server/form.php
MatthewMeacham: Beautiful 🥇

assets/js/signup.js
MatthewMeacham: I think a config file might be too fragile because how will you specify which question goes in which order? We can't do it by an id since those are subject to change.. I guess we could do it by the tag, but that seems pretty wonky to me. "Check this file every time we load the appointment sign up page, and sort by whatever this file says." <- That seems janky, it would be far easier to specify it in SQL and do a simple ORDER BY index ASC.

JakeKo: How would we handle needing to add a question or a subheading in the middle of the ordering? Or are we assuming the questions and subheading will never change?

MatthewMeacham: Pretty simple change the indices. Programmaticly, it wouldn't be hard. Obviously harder than a config file, but a SQL index would be far easier on the PHP side when working with it.

assets/js/signup.js
MatthewMeacham: We can't just put everything in a config file @JakeKo

JakeKo: I think you're referring to my comment immediately above this one which is a repeat of my comment in the earlier thread. Not sure why it's repeated.

CollinsSpencer:
Another option to help simplify front-end:

After looking at documentation for jQuery validation more, I think it might be best to use the [```rules()```](https://jqueryvalidation.org/rules/) method. This required that the ```validate()``` method be called first, so it would be easiest to have the validation rules in a different table. This would also help solve the problem when a form input needs to have multiple validation methods on it.

Therefore, we could pull all of the validation methods out of the question and questionInformation tables and just make a questionValidation table. This table would likely have
* a self-incrementing primary key id
* a required foreign key to the question (to get the tag, which will match the css id)
* a required validation type
* a required validation value
* an optional validation message

Most of the validation values will just be ```true```, but in the case of validation types like min, max, step, etc. it could be a number, so I'm not sure what would work best for the database column type for the validation value.

Pros:
* It greatly simplifies the code in signup.js and form.js
* It allows for custom error messages based on multiple validation types for the same question
* It lets us use validation methods like minlength, min, or step
* 2 fewer columns in the questionInformation table

Cons:
* New table
* It doesn't solve the issue of an invalid validationType in the table

PR #42, opened 2017-05-17
Redo questions db

MatthewMeacham:
Redid the questions DB after the talk with Austin. Basically, the Questions table was only meant to store the litmus test questions, not stuff like "What is your first name", "What is your last name", etc. This means that the litmus test questions are all select inputs. So the database has been redesigned to support that instead of what we were doing before.
- Renamed Question table to LitmusQuestion in order to avoid any confusion in the future about what type of questions this table is supposed to hold.
- Created a Client table which stores a client's firstName, lastName, emailAddress, and a nullable languages section which would store the languages they know if they're not fluent in English
- Both LitmusQuestion and PossibleAnswer now have a "orderIndex" field which specifies the ordering in which they should appear on the form and in the select input, respectively. A simple ORDER BY orderIndex ASC will get those in the proper order.
- Left PossibleAnswer table which stores possible answers for a question (the options in a select input)
- Left LitmusQuestion table which stores the questions
- Answer table now references an Appointment, a LitmusQuestion, and a PossibleAnswer. From that you can determine what the answers were for any appointment.
- Fixed Queue after this change
- Fixed test data

Here is the current, working ERD:

appointmentdberd

### DIFF
--- a/mysql scripts/table_creation/appointmentDB.sql
+++ b/mysql scripts/table_creation/appointmentDB.sql
@@ -8,10 +8,10 @@ DROP TABLE IF EXISTS Appointment;
 DROP TABLE IF EXISTS Client;
 DROP TABLE IF EXISTS Location;
 DROP TABLE IF EXISTS PossibleAnswer;
-DROP TABLE IF EXISTS Question;
+DROP TABLE IF EXISTS LitmusQuestion;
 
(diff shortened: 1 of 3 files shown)


2mnyzs:
:+1: 

Sorry about the misunderstanding.

Issue #2, opened 2017-02-21
Appointment Signup


2mnyzs:
- Users need to be able to signup
- Name, email, phone, address?
- Litmus test of questions to determine return complexity
    - Do you sell drugs?
    - Are you military?
    - Do you gamble?
    - Etc
    - User Litmus questions to also determine preparer cert requirements
- Language requirements
    - English, Spanish, C++, etc
- Choose site, then day, then time
- Gets an email confirming appointment
- Adds it into database
- Should be stored on the >queue

MatthewMeacham:
Closing this since this is basically the PR we just closed.

PR #47, opened 2017-06-19
Appointment Signup #2

CollinsSpencer:
The appointment signup page is now created entirely on the back-end. The litmus questions are the only dynamically added parts. 

On submit, the form will be validated before going to the server at all. If valid, it will be sent to the server as JSON formatted like `{1: "1", 2: "1", 3: "1", 4: "1", 5: "1", 6: "1", 7: "1", 8: "2", 9: "1", 10: "1", 11: "1", firstName: "First", lastName: "Last", email: "testName@example.com", phone: "4023331122"}` where the keys are the `litmusQuestionId` entries from `LitmusQuestion` and the values are the `orderIndex` entries from `possibleAnswers`.

On success, the page is redirected to the home page.  We will probably want to redirect to a page that says that their form was successfully submitted, tell them to check their email for a verification, etc.


MatthewMeacham:
Instead of using the orderIndex from PossibleAnswer you should probably just use the PossibleAnswer id. That makes more sense to me especially since the order index could potentially change in the future.

server/signup.php
```diff
@@ -1,12 +1,50 @@
...
+      	ORDER BY orderIndex');
+  $questionStatement->execute();
+  $questions = $questionStatement->fetchAll();
+
+  $optionStatement = $conn->prepare('SELECT possibleAnswerId, string, orderIndex from PossibleAnswer
+        WHERE litmusQuestionId = ?
+        AND archived = FALSE
+        ORDER BY orderIndex');
+
+  foreach ($questions as $question) {
```
MatthewMeacham: It might be better to perform these two SQL statements all in one rather than getting the questions and then getting the options for each question in the for loop. It's the difference between 1 database call and 11+. Something like:

SELECT lq.litmusQuestionId, pa.possibleAnswerId, pa.string AS possibleAnswerText, pa.orderIndex AS possibleAnswerOrderIndex,
	lq.string AS litmusQuestionText, lq.orderIndex AS litmusQuestionOrderIndex, lq.required AS required
FROM PossibleAnswer pa
JOIN LitmusQuestion lq ON pa.litmusQuestionId = lq.litmusQuestionId
WHERE lq.archived = FALSE AND pa.archived = FALSE
ORDER BY lq.orderIndex

Then sort it out into a dictionary in the PHP. That way we only do 1 database call.

MatthewMeacham: It might be better to perform these two SQL statements all in one rather than getting the questions and then getting the options for each question in the for loop. It's the difference between 1 database call and 11+. Something like:

SELECT lq.litmusQuestionId, pa.possibleAnswerId, pa.string AS possibleAnswerText, pa.orderIndex AS possibleAnswerOrderIndex,
	lq.string AS litmusQuestionText, lq.orderIndex AS litmusQuestionOrderIndex, lq.required AS required
FROM PossibleAnswer pa
JOIN LitmusQuestion lq ON pa.litmusQuestionId = lq.litmusQuestionId
WHERE lq.archived = FALSE AND pa.archived = FALSE
ORDER BY lq.orderIndex

Then sort it out into a dictionary in the PHP. That way we only do 1 database call.

MatthewMeacham: It might be better to perform these two SQL statements all in one rather than getting the questions and then getting the options for each question in the for loop. It's the difference between 1 database call and 11+. Something like:

SELECT lq.litmusQuestionId, pa.possibleAnswerId, pa.string AS possibleAnswerText, pa.orderIndex AS possibleAnswerOrderIndex,
	lq.string AS litmusQuestionText, lq.orderIndex AS litmusQuestionOrderIndex, lq.required AS required
FROM PossibleAnswer pa
JOIN LitmusQuestion lq ON pa.litmusQuestionId = lq.litmusQuestionId
WHERE lq.archived = FALSE AND pa.archived = FALSE
ORDER BY lq.orderIndex

Then sort it out into a dictionary in the PHP. That way we only do 1 database call.

PR #48, opened 2017-06-23
Added all table creation to FullDB.sql

MatthewMeacham:
Wrote all the SQL for creating all the necessary tables. The final ERD looks like this: 
Also @CollinsSpencer Don't hate me, but I changed the "string" column on LitmusQuestion and on PossibleAnswer to be "text" instead. I didn't like string.

I will write up test data in a separate issue.

And here is the ERD, note that audit trail connections back to the User table have been hidden.

databaseerd


mysql scripts/table_creation/FullDB.sql
```diff
@@ -0,0 +1,173 @@
...
+CREATE TABLE LitmusQuestion (
+	litmusQuestionId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
+    text VARCHAR(255) NOT NULL,
+	orderIndex INTEGER NOT NULL,
+    tag VARCHAR(255) NOT NULL,
+    required BOOLEAN NOT NULL DEFAULT TRUE,
+    archived BOOLEAN NOT NULL DEFAULT FALSE,
+    followUpTo INTEGER UNSIGNED NULL,
+    FOREIGN KEY(followUpTo) REFERENCES Litmusquestion(litmusQuestionId),
+	CONSTRAINT uniqueTag UNIQUE INDEX(tag)
```
JakeKo: Should be tabs IMO but nothing is written down anywhere....

mysql scripts/table_creation/FullDB.sql
```diff
@@ -0,0 +1,172 @@
...
+
+CREATE TABLE UserPrivilege (
+	userPrivilegeId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
+    userId INTEGER UNSIGNED NOT NULL,
+    FOREIGN KEY(userId) REFERENCES User(userId),
+    privilegeId INTEGER UNSIGNED NOT NULL,
+    FOREIGN KEY(privilegeId) REFERENCES Privilege(privilegeId)
+);
+
+CREATE TABLE Ability (
```
JakeKo: Might we consider a description or is that unnecessary if the tag is something like SpeaksSpanish? Or is that what name is supposed to be?

mysql scripts/table_creation/FullDB.sql
```diff
@@ -0,0 +1,172 @@
...
+	preparesTaxes BOOLEAN NOT NULL DEFAULT FALSE, 
+	archived BOOLEAN NOT NULL DEFAULT FALSE
+);
+
+
+CREATE TABLE ServicedAppointment (
+	servicedAppointmentId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
+    startTime DATETIME,
+    endTime DATETIME,
+    createdAt TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
```
JakeKo: Do we need createdAt? Until we figure out how much of an audit trail we need I don't know how much we want to half-record this information.

JakeKo: This might also be redundant if Appointment already has a createdAt field.

mysql scripts/table_creation/FullDB.sql
MatthewMeacham: My b, some of this was copy and pasted from the old version. I'll tabify it all

mysql scripts/table_creation/FullDB.sql
MatthewMeacham: A createdAt records when the record was actually inserted into the database. So no, this is not redundant since an Appointment createdAt would be set when the client makes an appointment. And the servicedAppointment createdAt would be set when a volunteer starts with a client. That being said, it is potentially slightly redundant since it would be created at either the startTime or the endTime in reality. Will change. And actually, on a side note, I'll look through each table and decide how much of an audit trail we need on each.

mysql scripts/table_creation/FullDB.sql
MatthewMeacham: I figured name would normally suffice, since it'll always be something like "Can Speak Spanish" or "Basic Certified". But we can totally add a description, I see no harm in doing so.

JakeKo: I'm a proponent of removing name entirely and replacing it with description and doing the same to the privileges table.

MatthewMeacham:
@CollinsSpencer The number of clients that can be serviced during a shift is a function of the number of volunteers working. The only flaw in your understanding is that appointments don't last 2 hours or however long a shift is. One volunteer can service multiple clients in a single shift. For v1, we plan on having X (number of volunteers) time slots available every 30 minutes.

2. It happens. We can't plan on that stuff in advance, but we do anticipate giving people the ability to change the number of "active, on-site volunteers" somewhere on our pages, so that if someone tries to make an appointment day-of, we can stop that if the number of volunteers is lower. Other than that, the site will just run behind. ¯\_(ツ)_/¯

mysql scripts/table_creation/FullDB.sql
MatthewMeacham: From a UI perspective, I believe that would be detrimental. On the webpage, would you rather see "volunteer has their basic certification" or "basic certification"? The first one was a description and the second was the name. It's even more crucial on the Privilege table, where some name may not be completely descriptive of what the privilege is. Take for example, "Can view suppliers". idk what that means, but with the description "Has permission to view all the suppliers on the /suppliers page." It's a contrived example, but we might as well document in the database rather than having a spreadsheet somewhere. Further, the description can be the hover effect on an item when someone hovers over it for more information.

PR #51, opened 2017-07-06
Create basic global styles

CollinsSpencer:
Due to the general lack of conformity for styling on the site so far, I have created a global style sheet. I separated up some of the more specific aspects like headings and buttons so that it is easier to jump into and make changes to in the future.

I know we might change fonts or sizing things in the future, but I figure this is good place to start.


MatthewMeacham:
:fire: :fire: :fire: :fire:

PR #53, opened 2017-07-07
Nav bar

CollinsSpencer:
This is a prototype of a nav bar. Obviously styling can be changed later. The main thing to discuss with this is how it is added to each page. ATM, is is added after the footer import. The reason for this is that the nav.php file includes a nav.js import. This js file relies on jQuery, which is imported in the footer.

While this makes the nav bar still appear at the top of the page like normal, it is located later in the DOM, which makes it load after the page itself. Probably the only page that won't have the nav bar in it is the login page, so we very well could just import the nav.js file in the footer. I'd rather not have to put the nav.js file in each file; it makes more sense to just include it in one of the other files that are included in almost every page.


MatthewMeacham:
LGTM. And while it would be a little confusing, I think you're right about just importing it in the footer since that will (presumably) be on every page. I don't think it's a bad thing to have this nav bar on every page either. And if it ever becomes a problem, we could always just do some if (/someWebPagePart/.test(window.location.href) and if that's true set the visibility of the nav bar to hidden lol.

assets/css/nav.css
```diff
@@ -0,0 +1,87 @@
...
+header#vita-nav nav a {
+  line-height: 3rem;
+  height: 3rem;
+  display: block;
+  text-decoration: none;
+  color: #ddd;
+  padding: 0 10px;
+}
+
+header#vita-nav nav.closed a + a { /* Using this as opposed to :not(:first-child) because it has support back through CSS2 */
```
MatthewMeacham: this is cool

index.php
```diff
@@ -4,10 +4,13 @@
 	require_once 'server/header.php';
 ?>
 	<body>
-	
-		<script src="assets/js/main.js"></script>
+		<div>
+			VITA Home Page
+		</div>
+		<!-- TODO -->
```
MatthewMeacham: what is this TODO for?

CollinsSpencer: I just put some random text on the page so we could see the different pages when you click between them in the nav bar. It was more just a comment that it won't say "VITA Home Page". And I guess just the fact that the whole page is a TODO. lol

CollinsSpencer:
Check out http://www.goodui.org/#14. I think I'm going to try changing the Litmus questions from being select inputs to actually being more like inline radio buttons without with little dots. I think that fits perfectly in this PR.

PR #54, opened 2017-07-07
Appointment signup fix

CollinsSpencer:
I updated the styling for the form to use some of the global styles. I also removed styling from form.css that was unique to the signup form.

I also found a couple spots where we missed the switch from string to text. I made those changes on this PR because it kind of fits.


MatthewMeacham:
Oh my bad on missing some of those string to texts and low-key don't use the production_data SQL one, made that a little prematurely. The test data one however will be kept up to date.

CollinsSpencer:
Check out http://www.goodui.org/#14. That was part of the reason that I changed the litmus questions. Plus, I found that exact functionality when I was looking into more jQuery UI stuff.

I also removed a redundant jQuery import. I ended up changing the import that we did have for jQuery to one that was one the same site as all of the rest of the other cdns.

assets/css/buttons.css
```diff
@@ -14,13 +14,13 @@
   font-weight: 400;
   font-style: normal;
   line-height: 1;
-  font-size: .75rem;
+  font-size: .75em;
```
JakeKo: What's the justification for using em? Particularly with font-size. I didn't catch it before but why not just px? Setting the font-size leads to ambiguity in what your font size truly is.

CollinsSpencer: I've been doing a bunch of reading into font sizing the past couple of days, and people on the internet are really conflicted and opinionated. My initial reason I made it em (or rather rem) is because of our intended audience. I didn't know if you guys thought that the font size should be larger. If you wanted it to be larger, then I would just change the font size for the html tag in one place and everything would change accordingly. I don't have much of an opinion either way, so if you think just using pixels makes more sense, then I could easily just change it over to that. Just let me know what you think.

JakeKo: Good point... It depends on whether we want every font/relevant elements to scale up at the same ratio. Some elements may not or should not scale up or down. I don't think we need to scale relatively but if that's our use case we should use em. Otherwise, px.

CollinsSpencer: For sure. I'm just not sure what makes sense to scale up or down. I know the problem with pixels is that they don't scale based on browser settings. Some people who have poor eye-sight set their browsers to have a default font size that is larger, which doesn't change the scale for how large "pixels" are. That's the difference between the browser's zoom feature and the accessibility font-size feature. Everything scales with the zoom. The problem I'm seeing is for mobile users, where the zoom feature isn't an option. They can zoom in, but it's more like a magnifying glass. Some people don't know how to turn on the larger text accessibility feature (which is ironic). That makes me wonder if on mobile there should be an button that can make the text larger right in the page. Reasonably, that would only affect the font-size on the page, but maybe not the nav bar? Just an idea. Not a necessity though. We probably need to focus on the features that allow volunteers to actually use our site first though. I'll put in an issue for this, and then we can worry about it more later.

2mnyzs:
I think I resolved the conflicts correctly, should be good now but would appreciate if someone could look over it.

I'm kind of curious why there's commented out css in the bottom of form.css but looks good other than that.
 
+1

PR #59, opened 2017-07-30
Queue

JakeKo:
We take things and then queue them.


mysql scripts/test_data/FullDBTestData.sql
```diff
@@ -245,6 +245,106 @@ INSERT INTO Appointment (scheduledTime, clientId, siteId)
...
+SET @appointmentTime = DATE_ADD((SELECT startTime FROM Shift WHERE shiftId = @shift_site2Shift2Id), INTERVAL 30 MINUTE);
+INSERT INTO Appointment (scheduledTime, clientId, siteId)
+	VALUES (@appointmentTime, @client_client4Id, @site_site2Id);
+SET @appointment_appointment4Id = LAST_INSERT_ID();
+
+SET @appointmentTime = DATE_ADD((SELECT startTime FROM Shift WHERE shiftId = @shift_site2Shift2Id), INTERVAL 30 MINUTE);
+INSERT INTO Appointment (scheduledTime, clientId, siteId)
+	VALUES (@appointmentTime, @client_client4Id, @site_site2Id);
+SET @appointment_appointment4Id = LAST_INSERT_ID();
+
```
MatthewMeacham: should we push all these duplicates?

queue/index.php
```diff
@@ -1,46 +1,65 @@
...
         <!--[if lt IE 8]>
             <p class="browserupgrade">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
-        <![endif]-->
-
-		<div class='vita-header'>
-			<div class='vita-volunteer-count'>Volunteers on duty: 0</div>
-			<img class='vita-header-curtain' src='../assets/res/queue_curtain.svg'>
-			<img class='vita-exit-fullscreen' src='../assets/res/generic_exit.svg'>
+		<![endif]-->
+		<?php require_once '../components/nav.php' ?>
```
MatthewMeacham: is there any reason this happens down below the ie stuff, or can this happen up by the other php require_once?

mysql scripts/test_data/FullDBTestData.sql
JakeKo: It's test data so I no reason why getting rid of them. It's nice to do some realistic load testing. Unless you want to make custom appointments for each of them - I'm not going to touch them.

server/queue.php
```diff
@@ -1,27 +1,24 @@
...
 	$conn = $DB_CONN;
+	$displayDate = $_GET['displayDate'];
 
 	// TODO make this handle multiple locations, if necessary
-	$stmt = $conn->prepare('SELECT DISTINCT appointmentId, scheduledTime, firstName, lastName
+	$stmt = $conn->prepare("SELECT DISTINCT appointmentId, scheduledTime, firstName, lastName
 		FROM Appointment
 		JOIN Client ON Appointment.clientId = Client.clientId
-		WHERE (Appointment.scheduledTime >= NOW() AND Appointment.scheduledTime < DATE_ADD(CURDATE(), INTERVAL 1 DAY))
+		WHERE DATE(Appointment.scheduledTime) = '$displayDate'
```
MatthewMeacham: correct me if I'm wrong, but this is screaming SQL injection to me. You probably shouldn't directly insert displayDate into this query. You should use the prepared statement to insert a dynamic variable at the execute part

assets/js/queue.js
```diff
@@ -1,29 +1,82 @@
...
+function keepTime() {
+	var h = new Date().getHours(),
+		m = new Date().getMinutes();
+
+	$('.clock-am').toggleClass('inactive-period', h >= 12);
+	$('.clock-pm').toggleClass('inactive-period', h < 12);
+
+	h %= 12;
+	if (h === 0) h = 12;
+	if (m < 10) m = `0${m}`;
```
MatthewMeacham: all this string interpolation is sexy

queue/index.php
JakeKo: This line brings in content so it needs to be in the body.

mysql scripts/test_data/FullDBTestData.sql
MatthewMeacham: I'd rather throw in a cursor or something than copy and pasting a bunch of the same appointment. I'll get on this in a separate PR, in the meantime, revert this file to what's on dev

queue/index.php
MatthewMeacham: ohhh got it

PR #60, opened 2017-07-30
Refresh database scripts

MatthewMeacham:
Created a powershell script to automatically run the create table and insert test data SQL scripts for developer convenience. 

NOTE that one will probably have to change the connection variables (namely username/password) if they have different information than what is in the config.php file


CollinsSpencer:
https://dev.mysql.com/downloads/connector/net/

PR #67, opened 2017-08-01
Privilege Checking Helper

2mnyzs:
Small helper class to make checking privileges easier. To test, you'll have to login which requires a user with associated login row. You'll to manually enter these if you don't have a local SMTP setup to register via email. 

Notes
1. Please note that privilege tags are defined constants, feel free to debate this. Should make it harder to mistype privileges. 
2. As Matt suggested, tags are used, not ids.
3. getUserId has a potentially unwanted side effect, will logout users past the threshold

### DIFF
--- a/server/user.class.php
+++ b/server/user.class.php
@@ -3,18 +3,13 @@ require_once 'login.class.php';
 require_once 'callbacks.php';
 require_once 'config.php';
 
-// META - Can give users permissions(inclusive?!)
 define('PrivilegeControl', 'PrivilegeControl');
-
-// Can Unassign users from a shift
 define('UserShiftControl', 'UserShiftControl');
-
-// Can Modify Appointments
 define('AppointmentControl', 'AppointmentControl');
 
 
 /**
-* Class for easily checking user permissions
+* Class for easily checking user privileges
 * Simply instantiate and use
 */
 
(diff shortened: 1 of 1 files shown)


server/user.class.php
```diff
@@ -0,0 +1,77 @@
...
+		$stmt = $DB_CONN->prepare($query);
+		$stmt->execute(array($userId,$privilegeTag));
+
+		$results = $stmt->fetchAll();
+
+		if(empty($results)){
+			return false;
+		}else{
+			return true;
+		}
```
MatthewMeacham: couldn't this just be ```return empty($results)`;``

server/user.class.php
```diff
@@ -0,0 +1,77 @@
...
+	*/ 
+	public function hasPrivilege($privilegeTag){
+		GLOBAL $DB_CONN;
+
+		$userId = $this->getUserId();
+
+		$query = "SELECT userId 
+			FROM userPrivilege
+				INNER JOIN privilege ON privilege.privilegeId = userprivilege.userPrivilegeId
+			WHERE 1=1
```
MatthewMeacham: Also, I still wanna hear your works reasoning on why you always do the 1=1?

server/user.class.php
2mnyzs: Oh definitely yea, I had more in mind when I wrote the if statement but ended up not needing it. 

Removed.

server/user.class.php
2mnyzs: Ofc, I don't mind removing it at all if you want

MatthewMeacham: No that all makes total sense to me, I was just curious

server/user.class.php
```diff
@@ -0,0 +1,73 @@
...
+define('AppointmentControl', 'AppointmentControl');
+
+
+/**
+* Class for easily checking user privileges
+* Simply instantiate and use
+*/
+
+class User
+{
```
JakeKo: Unforgivable inconsistency among brace placement is unforgivable

PR #68, opened 2017-08-05
Added test data for each different kind of user

MatthewMeacham:
We needed test data for users so that we can begin testing the logged in / permissions functionality. You can login by navigating to /login and then using a valid email/password. Then if successful, it will redirect you back to the index page (this will eventually be changed once other pages are up and running).

And here are the email/password for each account. 

| email                  | password |
|------------------------|----------|
| preparer@test.test     | test     |
| reviewer@test.test     | test     |
| receptionist@test.test | test     |
| siteadmin@test.test    | test     |


2mnyzs:
+1 Please merge.

PR #69, opened 2017-08-06
Large Private Queue Update

JakeKo:
This PR adds a lot to the private queue. At this point in time, the private queue is highly functional. A volunteer is able to view the up-to-date queue, select an appointment, and inspect the details for that appointment. They can also close the appointment without cancelling it (closing simply hides the appointment details) and they can cancel the appointment.

A volunteer can NOT accept or reschedule an appointment yet.

A grievance:
* The page will refresh every 10 minutes. This is not good because that will end whatever appointment the volunteer is in.


assets/js/queue_priv.js
```diff
@@ -0,0 +1,104 @@
+var REFRESH_SEC = 15,
+	displayDate = new Date(),
+	monthStrings = [ 'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec' ],
```
MatthewMeacham: The amount of code duplication between this js and the queue.js is really unfortunate, is there any way we could have a shared js for anything that is shared between them, and then separate files for the things that are different? Or is that not gonna be able to work in this instance?

JakeKo: I updated my PR overview comment "explaining" the situation.

MatthewMeacham: :fire: got it thanks bro

MatthewMeacham:
Just some more stuff to think about when you make the next update for this page.

It might be nice to have whatever appointment they're clicked on in the left to be like shaded or highlighted or whatever one calls it, just so they can more easily know which one they clicked on without having to compare names. 

Additionally, a 'legend' for what the NS and OT means might be nice (ik they mean no show and on time, but others might not understand that. OT does tend to mean overtime, especially to sportsball fans, so a legend would avoid this ambiguity)

Also rather than that close button, would an X at the top right of the appointment details box be more intuitive?

Also a volunteer probably only needs to see the appointments at their site (with the ability to see all appointments if they want). So idk how we want to implement this functionality -- once a volunteer is logged on, it finds what shift they are at right, and then finds the site for that shift? And that is the site it shows, plus a select for them to choose a site or all sites? Maybe?

And then eventually, having the if (logged in) stuff happening so that the only URL for the queue is /queue rather than both /queue and /queue_priv

MatthewMeacham:
LGTM

assets/js/queue_common.js
```diff
@@ -0,0 +1,52 @@
+var displayDate = new Date(),
+	monthStrings = [ 'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec' ];
+
```
2mnyzs: We could consider using a library like Moment for stuff like this, looks fine though.

PR #70, opened 2017-08-14
Appointment signup fix

2mnyzs:
Added a script and modified the JS to store appointments. Sends an email notification upon completion and notifies the client upon success.

To test:
1. Go to /signup/
2. Sign up
3. Look at confirmation message
4. Make sure the database has data in it


assets/js/signup.js
```diff
@@ -1,101 +1,92 @@
...
+		rules: {
+			"firstName": "required",
+			"lastName": "required",
+			email: {
+				required: true,
+				email: true
+			},
+			phone: {
+				required: true,
+				// phoneUS: true
```
MatthewMeacham: Was this commented out since we _can_ have international students? Is that the thought process here?

assets/js/signup.js
```diff
@@ -1,101 +1,92 @@
...
-  '  }' +
-  '}';
+	// AJAX Code To Submit Form.
+	$.ajax({
+		url: "/server/storeAppointment.php",
+		type: "post",
+		dataType: 'json',
+		data: (data),
+		cache: false,
+		complete: function(response){
```
MatthewMeacham: why use completed and not use like failure and success? I feel like those would be more idiomatic than this complete one.

server/config.php
```diff
@@ -1,5 +1,5 @@
 <?php
-	define('PROD', false);
+	define('PROD', true);
```
MatthewMeacham: uh did we want this change to go through?

server/storeAppointment.php
```diff
@@ -0,0 +1,120 @@
...
+
+		// TODO: Make sound better
+		$response['message'] = $data['firstName'].", thank you for signing up! Your appointment will be located at $siteAddress. Please arrive by ".$data['scheduledTime']." with all necessary materials. Please call $sitePhoneNumber for additional details or to reschedule. Thank you from Lincoln VITA.";
+
+		// if email is set and passes simple validation (x@x)
+		if($data['email'] && preg_match('/.+@.+/', $data['email'])){
+			mail($data['email'], 'Lincoln VITA - Appointment', $response['message']);
+		}
+	} catch (Exception $e) {
+		$conn->rollback();
```
MatthewMeacham: should we really rollback if they entered an invalid email address _somehow_? I feel like a rollback isn't necessary. Presumably the rest of the data is valid-ish? Then again, I suppose if someone is giving data not through the form (or at least bypassing JS validation), then that's _probably_ malicious? Further, the person would never know that their appointment was essentially cancelled since they presumably would've gotten the message above saying "we can't wait to see you!" or whatever it says. So that's not cool either.

server/storeAppointment.php
```diff
@@ -0,0 +1,120 @@
+<?
+require_once 'config.php';
+
+storeAppointment($_POST);
+
+function storeAppointment($data){
+	GLOBAL $DB_CONN;
+	$conn = $DB_CONN;
```
MatthewMeacham: btw what is the reasoning for creating a new variable to hold the connection, when DB_CONN already has it?

assets/js/signup.js
2mnyzs: Forgot to comment on this.

You have to include a separate lib to use jquery phone validation(for some reason)
https://stackoverflow.com/questions/16652776/jquery-validation-plugin-phoneus-field-is-breaking-form#answer-16652933

And in general I try to avoid over-validating(as I see it) input that doesn't really need validation. Maybe make sure there are at least 10 numbers but I feel that anything beyond that degrades UX needlessly.

assets/js/signup.js
2mnyzs: Could do that but I'm not sure on what to do on the `error` callback? Tell them to try again with better internet?

server/config.php
2mnyzs: Whoops, reverted

server/storeAppointment.php
2mnyzs: We should probably talk about this, I'm unsure of what you're getting at.

server/storeAppointment.php
2mnyzs: Holdover from something, changed to use the global.

server/storeAppointment.php
MatthewMeacham: whoops my bad, thought this try was only around the mailer part. Never mind

assets/js/signup.js
```diff
@@ -1,101 +1,92 @@
...
+	});
 }
 
 // Form submission
 $('#vitaSignupForm').submit(function(e) {
-  if (!$(this).valid()) {
-    return false;
-  }
+	// Stop default form submit action
+	e.preventDefault();
```
CollinsSpencer: This line redundant. In jQuery, return false; effectively calls e.preventDefault and e.stopPropagation.

See [StackOverflow](https://stackoverflow.com/questions/1357118/event-preventdefault-vs-return-false) for more.

assets/js/signup.js
CollinsSpencer: We probably should be using the success callback. I don't think we would want to treat errors as being successful.

PR #86, opened 2017-08-31
Refactor directories to more logically group files

CollinsSpencer:
Note, nav.js and nav.css were not pulled out of the assets/css folder. I consider these files to be global.


MatthewMeacham:
/signup gives a `GET http://localhost/assets/js/signup.js 404 (Not Found)`
/login does as well for the `login.js`
/register gives a not found for 'login.js'

Other than that, looks good, all styling seems to work, and nicely done the queue portion, that is what I had in mind as well (once we get the signed in logic in)

Issue #85, opened 2017-08-31
The sign up process needs to be redone to account for the different paths a person could take.


MatthewMeacham:
Currently, the signup page is just a contact information + a few litmus questions. This needs to be redone to account for the different types of taxes that get done at VITA sites (residential, international, military, etc.). Please refer to the flowchart to see how the residential/international are handled. 

Some other important questions to ask:
* we need to ask if the person is a UNL student because the center needs to keep track of this information for reporting purposes.
* another important question that they want is to know "if there will be multiple years of returns" as something like this requires
* I believe they also mentioned something about the number of tax returns actually happening in the appointment. i.e. if some guy brings in his family of 5 and they all have tax returns to do and they only had a single appointment, that would put us behind. So knowing the total number of returns beforehand would be helpful.

Spend some time thinking about the different information that we may need for a client, and make the transition for questions as smooth as possible. Something like this can probably just be a hard-coded front-end rather than trying to pull questions / question branches from the database.

PR #94, opened 2017-09-02
Database changes

MatthewMeacham:
Renamed some tables and columns and removed some columns. See comments for reasoning.

databaseerd


mysql_scripts/table_creation/FullDB.sql
```diff
@@ -33,12 +38,12 @@ CREATE TABLE LitmusQuestion (
 	litmusQuestionId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
 	text VARCHAR(255) NOT NULL,
 	orderIndex INTEGER NOT NULL,
-	tag VARCHAR(255) NOT NULL,
+	lookupName VARCHAR(255) NOT NULL,
```
MatthewMeacham: tag is wildly ambiguous in my opinion, switching to lookupName because it clearly describes that this field is used for lookups in this table based on a string value.

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -33,12 +38,12 @@ CREATE TABLE LitmusQuestion (
...
 	required BOOLEAN NOT NULL DEFAULT TRUE,
 	archived BOOLEAN NOT NULL DEFAULT FALSE,
 	followUpTo INTEGER UNSIGNED NULL,
 	FOREIGN KEY(followUpTo) REFERENCES Litmusquestion(litmusQuestionId),
```
MatthewMeacham: I'm leaving alone the entire question system. This will be dealt with in #85

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -129,32 +134,32 @@ CREATE TABLE PasswordReset (
 	FOREIGN KEY (userId) REFERENCES User(userId)
 );
 
-CREATE TABLE Privilege (
-	privilegeId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
+CREATE TABLE Permission (
```
MatthewMeacham: I didn't like the term privilege either. It doesn't make as much sense as permission in my opinion.

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -172,9 +177,7 @@ CREATE TABLE Shift (
 	shiftId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
 	startTime DATETIME NOT NULL,
 	endTime DATETIME NOT NULL,
-	handlesResidential BOOLEAN NOT NULL,
-	handlesInternational BOOLEAN NOT NULL,
-	deleted BOOLEAN NOT NULL DEFAULT FALSE,
```
MatthewMeacham: Got rid of the handlesResidential / handlesInternational since this was a bad way to set it up since VITA handles 5 different kinds of taxes. A more robust system will have to be made to handle these situations where this site is only handling X taxes these days. But also, she said it's a hard date that they start doing international taxes (Tuesdays/Thursdays in March/April), so we can get away with hardcoding this for now

Renamed deleted to archived to be more standard across the rest of the system

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -129,32 +134,32 @@ CREATE TABLE PasswordReset (
...
+	permissionId INTEGER UNSIGNED NOT NULL,
+	FOREIGN KEY(permissionId) REFERENCES Permission(permissionId)
 );
 
 CREATE TABLE Ability (
 	abilityId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
 	name VARCHAR(255) NOT NULL,
-	tag VARCHAR(255) NOT NULL,
+	lookupName VARCHAR(255) NOT NULL,
     verificationRequired BOOLEAN NOT NULL,
```
CollinsSpencer: Lines 160, 167, 168, and 169 have spaces instead of tabs. Otherwise every thing looks good. I'll fix that really quick and merge it.

PR #95, opened 2017-09-02
Switch to Bootstrap 4

CollinsSpencer:
Bootstrap 4 uses flexbox now and is fantastic. It's really easy to use and looks very clean.

I've still got the queue and some other small things to convert to Bootstrap, but I figured I'd stop here because this is already getting to be a lot of changes.

When I convert the queue, I will also switch that over to basic AngularJS templating.


components/nav.php
```diff
@@ -10,15 +10,38 @@
...
+			</span>
+
+			<ul class="navbar-nav">
+				<li class="nav-item">
+					<a class="nav-link" href="/queue">Queue</a>
+				</li>
+				<li class="nav-item">
+					<a class="nav-link" href="/login">Volunteer Login</a>
+				</li>
+			<?php if (isset($_REQUEST['admin'])): ?>
```
CollinsSpencer: First of all, this is temporary. Throw `?admin` onto the url to see the additional menus.

server/footer.php
```diff
@@ -1,14 +1,20 @@
...
-        e.src='https://www.google-analytics.com/analytics.js';
-        r.parentNode.insertBefore(e,r)}(window,document,'script','ga'));
-        ga('create','UA-XXXXX-X','auto');ga('send','pageview');
-    </script>
+	<!-- AngularJS -->
+	<script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.6.4/angular.min.js"></script>
+	<!-- Bootstrap -->
+	<script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN" crossorigin="anonymous"></script>
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.11.0/umd/popper.min.js" integrity="sha384-b/U6ypiBEHpOf/4+1nzFpr53nxSS+GLCkfwBdFNTxtclqqenISfwAzpKaMNFNmj4" crossorigin="anonymous"></script>
+	<script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/js/bootstrap.min.js" integrity="sha384-h0AbiXch4ZDo7tp9hKZ4TsHbi047NrKGLO3SEJAg45jXxnGIfYzk4Si90RDIqNm1" crossorigin="anonymous"></script>
```
CollinsSpencer: Again, notice that this is Bootstrap 4. It makes use of flexbox and makes formatting a whole lot easier than past versions. So, for example, you can no longer do `col col-6 col-offset-3` to center a chunk that is half of the screen width. Now you can center with `justify-content-center` on the row div, `align-self-center` pm the column, or `mx-auto` on the column. This provides lots of opportunities when it comes to laying out components relative to one another.

Some useful pages from the Bootstrap docs:
* [Grid System](https://getbootstrap.com/docs/4.0/layout/grid/)
* [Colors](https://getbootstrap.com/docs/4.0/utilities/colors/)
* [Spacing](https://getbootstrap.com/docs/4.0/utilities/spacing/)

server/header.php
```diff
@@ -4,10 +4,11 @@
 <meta name="viewport" content="width=device-width, initial-scale=1">
 
 <link rel="apple-touch-icon" href="apple-touch-icon.png">
-<link rel="stylesheet" href='https://cdnjs.cloudflare.com/ajax/libs/normalize/6.0.0/normalize.min.css'>
+<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/css/bootstrap.min.css" integrity="sha384-/Y6pD6FV/Vv2HJnA6t+vslU6fwYXjCFtcEpHbNJ0lyAFsXTsjBbfaDjzALeQsN6M" crossorigin="anonymous">
 <link rel="stylesheet" href="/assets/css/global.css">
 <link rel='stylesheet' href='/assets/css/nav.css'>
 <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jqueryui/1.12.1/jquery-ui.min.css">
```
CollinsSpencer: Just as a note, some of these references might get pulled later.

PR #100, opened 2017-09-18
Added translation of integral languages

rthakker549:


MatthewMeacham:
Looks good and appears to work. My only concern is that this thing is pretty hard to see and also pretty hard to find. Idk if clients are going to be scrolling all the way to the bottom if they can't even read the first sentence of the site? Is there a better place / better way to display this?

CollinsSpencer:
Should work to just add this code to `assets/css/main.css`
```
#google_translate_element {
  position: fixed;
  bottom: 0;
}
```

MatthewMeacham:
Updates on this @rthakker549?

PR #101, opened 2017-10-08
Site Admin Download Appointment Schedule for Site

MatthewMeacham:
~~She's not very pretty, but I'd like to get some comments on the functionality here. Is there something we should do differently for these things? This is a pretty important feature that they want just as a fail-safe in case there's a huge bug in our system or whatever. So any input on how to improve this would be appreciated.~~

~~Also, input on how to make it look prettier is also appreciated. As well as file names, I didn't know what to call half of this stuff...~~

~~I think it would be nice to have it exported as an excel file so that we could have a different workbook for each site and then they could download all sites at once in one file (rather than 10 files). But I'm not sure if we can do this on the CSE server yet (it will require installing a new PHP module). If we can do this, then this will definitely be the way to go. But if we can't do that, then how could we make a CSV easier to use?~~

**UPDATE**: Updated this to use PHPExcel and create an excel file with a different sheet for each site. The user now has the option to download all sites at once or choose one individually for a single day.

You can test this functionality by:
* running the SQL files to generate test data (or running the powershell script)
* Logging in as a siteadmin using the credentials detailed in the wiki
* Making sure the nav bar has the Admin dropdown, and selecting the "Print documents tab"
* Selecting a date / site, and hitting download schedule
* Ensuring the excel file has the data that it should based on what is in the database

### DIFF
--- a/components/nav.php
+++ b/components/nav.php
@@ -1,4 +1,9 @@
 <!-- TODO: insert authentication logic here -->
+<?php 
+	$root = realpath($_SERVER["DOCUMENT_ROOT"]);
+	require_once "$root/server/user.class.php";
+	$USER = new User();
+?>
 
 <?php
 	if (!isset($page_title)) {
@@ -30,13 +35,13 @@
 				<li class="nav-item">
 					<a class="nav-link" href="/login">Volunteer Login</a>
 				</li>
-			<?php if (isset($_REQUEST['admin'])): ?>
+			<?php if ($USER->hasPermission('can_view_management_tab')): ?>
 				<li class="nav-item dropdown">
 					<a class="nav-link dropdown-toggle" href="#" id="dropdown01" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Admin</a>
 					<div class="dropdown-menu dropdown-menu-right" aria-labelledby="dropdown01">
 						<a class="dropdown-item" href="#">Review Certifications</a>
 						<a class="dropdown-item" href="#">Adjust Shifts</a>
-						<a class="dropdown-item" href="#">Print Documents</a>
+						<a class="dropdown-item" href="/management/appointments">Print Documents</a>
 					</div>
 				</li>
 			<?php endif; ?>
--- a/components/nav.php
+++ b/components/nav.php
@@ -1,11 +1,8 @@
-<!-- TODO: insert authentication logic here -->
-<?php 
+<?php
 	$root = realpath($_SERVER["DOCUMENT_ROOT"]);
 	require_once "$root/server/user.class.php";
 	$USER = new User();
-?>
 
-<?php
 	if (!isset($page_title)) {
 		$page_title = "VITA";
 	}
--- a/management/appointments/appointments.js
+++ b/management/appointments/appointments.js
@@ -26,7 +26,7 @@ let downloadCsv = function() {
 			exportAsCsv(csvString, fileName);
 		},
 		error: function(response) {
-			alert("Unable to get data");
+			alert("Unable to get appointments. Please refresh the page in a few minutes.");
 		}
 	});
 };
@@ -86,14 +86,13 @@ let loadAllSites = function() {
 		}),
 		cache: false,
 		success: function(response) {
-			console.log(response);
 			let siteSelect = document.getElementById("siteSelect");
 			for ($i = 0; $i < response.length; $i++) {
 				siteSelect.options.add(new Option(response[$i].title, response[$i].siteId));
 			}
 		},
 		error: function(response) {
-			alert("Unable to load sites");
+			alert("Unable to load sites. Please refresh the page in a few minutes.");
 		}
 	});
 }
--- a/management/appointments/appointments.js
+++ b/management/appointments/appointments.js
@@ -1,3 +1,7 @@
+$(document).ready(function() {
+	loadAllSites();
+});
+
 let defaultColumnNames = ["Scheduled Time", "First Name", "Last Name", "Phone Number", "Email Address", "Appointment ID"];
 
 let downloadCsv = function() {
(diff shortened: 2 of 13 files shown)


management/appointments/index.php
```diff
@@ -0,0 +1,40 @@
...
+		<select id="siteSelect">
+			<?php
+				require_once "$root/server/config.php";
+				GLOBAL $DB_CONN;
+				
+				$query = "SELECT siteId, title FROM Site";
+				$stmt = $DB_CONN->prepare($query);
+
+				$stmt->execute();
+				$sites = $stmt->fetchAll(PDO::FETCH_ASSOC);
```
CollinsSpencer: Just from a design standpoint, I'm not a fan of having back-end/database calls in a file that should be for the front-end/view. In this case I think it's probably fine because it's just grabbing the names of the sites, so it would probably be a bit extra to have a separate file.

CollinsSpencer:
For looks, it's not that bad. I would suggest you add some padding to the heading though. Just add a bootstrap class to the h2. The [new bootstrap spacing utilities](http://getbootstrap.com/docs/4.0/utilities/spacing/) make it really easy. For example you could add `pt-5` and it would give the heading some spacing that would look good.

management/appointments/index.php
MatthewMeacham: Oh yah I meant to pull this one out, thanks Spencer

MatthewMeacham:
@CollinsSpencer  I created an unauthorized page that will need to be made to look pretty sometime... I created #103 as an issue, just something that someone else should know about.

I created a server/api/sites/getAll.php to contain the functionality for getting sites, there's no filtering functionality on it, but this is a good basis for basic getting site information that is going to happen in other places. I think adopting this sort of api pattern would be nice to try to keep our /server directory cleaner and less scary-looking. I'm open to renaming these directories/moving them if anyone has a different opinion.

components/nav.php
```diff
@@ -30,13 +32,13 @@
 				<li class="nav-item">
 					<a class="nav-link" href="/login">Volunteer Login</a>
 				</li>
-			<?php if (isset($_REQUEST['admin'])): ?>
+			<?php if ($USER->hasPermission('can_view_management_tab')): ?>
```
2mnyzs: Haven't seen the if:endif syntax in awhile but I can't see why not.

management/appointments/appointments.js
```diff
@@ -0,0 +1,99 @@
...
+		cache: false,
+		success: function(response) {
+			console.log(response);
+			let siteSelect = document.getElementById("siteSelect");
+			for ($i = 0; $i < response.length; $i++) {
+				siteSelect.options.add(new Option(response[$i].title, response[$i].siteId));
+			}
+		},
+		error: function(response) {
+			alert("Unable to load sites");
```
2mnyzs: I usually put some instruction in error messages like: "Please refresh the page in a few minutes." Or "Check your internet connection and try again." Just so they have some idea of what to do next.

unauthorized/index.php
```diff
@@ -0,0 +1,20 @@
+<?php $root = realpath($_SERVER["DOCUMENT_ROOT"]) ?>
```
2mnyzs: I kind of like this idea. Another alternative is returning a HTTP 403 but I think this one is nice because we can end up using this page as a sort of readme on permissions like "please contact your site coordinator if you need access" or similar.

server/api/sites/getAll.php
```diff
@@ -0,0 +1,43 @@
...
+function readAllSites($data) {
+	GLOBAL $DB_CONN;
+
+	// construct select parameter list
+	$selectParameters = '';
+	if ($data) {
+		foreach (defaultSelectParameters as $key) {
+			if (isset($data[$key])) {
+				if ($data[$key] == true) {
+					$selectParameters = $selectParameters . ' ' . $key; // append the select parameter
```
2mnyzs: Is this being used for anything?

unauthorized/index.php
2mnyzs: Idea being the entire page.

server/api/sites/getAll.php
MatthewMeacham: It's supposed to be, good catch

management/appointments/appointments.js
MatthewMeacham: 👍 Good idea, will do

components/nav.php
MatthewMeacham: That was there when I got there. Is there a better way we should do it?

unauthorized/index.php
MatthewMeacham: I also think this is nice since this isn't exactly an API, but rather like a user facing site, an unauthorized thing looks nice and feels cleaner.

server/api/sites/getAll.php
```diff
@@ -0,0 +1,43 @@
...
+ *   "address": false
+ * }
+ * 
+ * If the database field is missing it is assumed that you DON'T want it. 
+ */
+function readAllSites($data) {
+	GLOBAL $DB_CONN;
+
+	// construct select parameter list
+	$selectParameters = '';
```
MatthewMeacham: Rename to selectColumns

MatthewMeacham:
Updated this to use PHPExcel. The code relies on the fact that PHPExcel is installed in the server/libs directory. I haven't committed those changes yet. What is the right way to do this? Is it to commit those files? Or is there a different way? @2mnyzs 

Once we get that figured out, this PR should be good to go.

MatthewMeacham:
Updated to use git submodules. You will need to run a `git submodule init` followed by a `git submodule update` in order for the PHPExcel submodule to be created

CollinsSpencer:
These new changes look good. No problems downloading each type.

Issue #75, opened 2017-08-30
As a client, I want to be able to select a time at any site for any day when signing up for an appointment.


MatthewMeacham:
Currently, the site id and appointment time are hardcoded in the signup.js when someone signs up for an appointment. Instead, a client needs to be able to select from a list of sites, and then select the date, and then see all the available times*. The times should be in increments of 1 hour.

For now, on this task, just don't do any filtering based on what sort of tax it is, we will handle that in another issue.

I don't know the best UX for this: should it be site -> date -> time or maybe date -> site -> time? Or maybe not a workflow like that at all? But bottom line, we need to get the site they want and the date/time they want it.

PR #102, opened 2017-10-08
Conditionals for Appoinment Signup

CollinsSpencer:
All of the questions are now hard-coded in the signup form and conditionals have been implemented.

Testing considerations:
- [x] check the conditionals that have subconditionals. e.g. question A has options "yes" and "no", question B has options "yes" and "no", and question C has options "yes" and "no". Only question A is shown by default. When "yes" is selected for question A, question B appears. When "yes" is selected for question B, question C appears. Then if question is is switched to "no" (hiding questions B and C) and then switched back to "yes", "yes" should still be selected for question B, and question C should still be shown.

Additional things:
- [x] I noticed that all of the questions that either are not shown or are not selected (and not required) will still send a value with undefined for their id and value back to the server. These aren't put into the database, but I didn't know if it would be better to just add a simple check in the javascript before sending it to the server at all. i.e. `if(checkedRadioBox.length>0)` as the check on line 183 in `signup.js`.

Notes:
* I did put in containers for the appointment signup portion of the form. These will be changed with issue #75.
* in storeAppointment.php, I commented out the lines that would send the emails. These will be changed later, but currently, they break the code and make the front-end think think there was an error.

### DIFF
--- a/server/storeAppointment.php
+++ b/server/storeAppointment.php
@@ -111,7 +111,7 @@ function storeAppointment($data){
 		$DB_CONN->rollback();
 
 		// TODO
-		// mail('someoneimportant@important.com', 'Please help, everything is on fire?', print_r($e, true).print_r($data, true));
+		// mail('vita@cse.unl.edu', 'Please help, everything is on fire?', print_r($e, true).print_r($data, true));
 	}
 
 	## Return
--- a/server/storeAppointment.php
+++ b/server/storeAppointment.php
@@ -105,13 +105,13 @@ function storeAppointment($data){
 
 		// if email is set and passes simple validation (x@x)
 		if($data['email'] && preg_match('/.+@.+/', $data['email'])){
-			mail($data['email'], 'Lincoln VITA - Appointment', $response['message']);
+			// mail($data['email'], 'Lincoln VITA - Appointment', $response['message']);
 		}
 	} catch (Exception $e) {
 		$DB_CONN->rollback();
 
 		// TODO
-		mail('someoneimportant@important.com', 'Please help, everything is on fire?', print_r($e, true).print_r($data, true));
+		// mail('someoneimportant@important.com', 'Please help, everything is on fire?', print_r($e, true).print_r($data, true));
 	}
 
 	## Return
--- a/signup/signup.js
+++ b/signup/signup.js
@@ -37,18 +37,132 @@ function validateSignupForm() {
 }
 
 function conditionalFormFields() {
-	// Independent fields = #student
-	var student = $("#student").find('input:radio[name="student"]');
-	// Dependent fields = #studentId
-	var studentId = $("#studentId");
-	var all = studentId;
-	// Hide everything and selectively show content
-	student.change(function(){
+	// All of the questions that required conditions to be viewed.
+	var homeBased = $("#homeBased");
+	var homeBasedNetLoss = $("#homeBasedNetLoss");
+	var homeBased10000 = $("#homeBased10000");
+	var homeBasedSEP = $("#homeBasedSEP");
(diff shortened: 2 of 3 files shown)


MatthewMeacham:
This is good stuff. Any chance you could make the page scroll down when someone hits a question at the bottom so that they don't have to scroll themselves?

MatthewMeacham:
Looks like it works correctly 👍

signup/signup.js
```diff
@@ -34,6 +36,137 @@ function validateSignupForm() {
...
+	var studenth1bValues = studenth1b.find('input:radio[name="18"]');
+
+	// To help hide everything and selectively show content
+	var allUnderHomeBasedValues = homeBasedNetLoss.add(homeBased10000).add(homeBasedSEP).add(homeBasedEmployees);
+	var allUnderstudentIntVisaValues = studentf1.add(studentj1).add(studenth1b);
+	var allUnderStudentIntValues = studentIntVisa.add(allUnderstudentIntVisaValues)
+	var allUnderStudentUNLValues = studentInt.add(allUnderStudentIntValues).add(allUnderstudentIntVisaValues);
+
+	// Independent field = #homeBased
+	// Dependent field = if yes --> #homeBasedNetLoss,#homeBased10000,#homeBasedSEP,#homeBasedEmployees
```
MatthewMeacham: Nice organizational pattern btw 🔥

signup/signup.js
```diff
@@ -34,6 +36,137 @@ function validateSignupForm() {
...
+	// Dependent field = if yes --> #studentInt
+	//                   if no --> appointmentPicker
+	studentUNLValues.change(function(){
+		var value = this.value;
+		allUnderStudentUNLValues.slideUp(300);
+		if(value === "1"){
+			studentInt.slideDown(300);
+			studentIntValues.change();
+		} else if(value === "2"){
+			// TODO
```
MatthewMeacham: What is the TODO here? What will we eventually do here?

MatthewMeacham:
Also, I think it would be best to add the check so we don't send the undefined/undefined pair to the backend. Although it gracefully handles it already, I think it would be better to be explicit.

server/storeAppointment.php
```diff
@@ -105,13 +105,13 @@ function storeAppointment($data){
...
 		if($data['email'] && preg_match('/.+@.+/', $data['email'])){
-			mail($data['email'], 'Lincoln VITA - Appointment', $response['message']);
+			// mail($data['email'], 'Lincoln VITA - Appointment', $response['message']);
 		}
 	} catch (Exception $e) {
 		$DB_CONN->rollback();
 
 		// TODO
-		mail('someoneimportant@important.com', 'Please help, everything is on fire?', print_r($e, true).print_r($data, true));
+		// mail('someoneimportant@important.com', 'Please help, everything is on fire?', print_r($e, true).print_r($data, true));
```
2mnyzs: I think we can use our new vita email for this: vita@cse.unl.edu. Thoughts?

MatthewMeacham: As mentioned in our phone call, I think that it's a good idea, but like you said it would be nice to have our main emails cc'd to also sent to, because chances are we won't be checking the vita email all the time. (Although I suppose we could set up like a POP3 or whatever and have it forwarded to us, then we wouldn't have to repeat a line of code). Perhaps a forwarding system is the way to go...

CollinsSpencer:
This should be GTG

MatthewMeacham:
This isn't something we need to deal with now (or maybe not even at all?). But if I say that I own a home-based business, and then answer the questions below that, and then go back and say that I don't own a home-based business, the answers to the subquestions still get sent back to the server and put in the database. I don't really think that affects anything, but it is something to keep in mind if it ever does become a problem.

PR #104, opened 2017-10-23
Permission(s) Control and Adding Users

2mnyzs:
Added a page: /manage/user that lets people with Edit_User_Permission to edit permissions as well as add new users to the system.

To test: a constraint was added to the permission table that prevents a permission for one user to appear more than once. Please add this constraint before testing, Then simply log in with the siteadmin@test.test:test user account and go to the manage user page. On the page simply click on a dropdown and then click a permission to add or remove it. Also try adding a user in the top right and then giving it some permissions.

A note on the css: the majority of it comes from the linked place to make the pretty bootstrap-select work with V4. I've noticed no issues with it but something to keep in mind.


mysql_scripts/table_creation/FullDB.sql
```diff
@@ -150,7 +150,8 @@ CREATE TABLE UserPermission (
 	userId INTEGER UNSIGNED NOT NULL,
 	FOREIGN KEY(userId) REFERENCES User(userId),
 	permissionId INTEGER UNSIGNED NOT NULL,
-	FOREIGN KEY(permissionId) REFERENCES Permission(permissionId)
+	FOREIGN KEY(permissionId) REFERENCES Permission(permissionId),
+	CONSTRAINT UNIQUE unique_permission (userId, permissionId)
```
MatthewMeacham: Good idea on this constraint

server/manageuser.php
```diff
@@ -0,0 +1,153 @@
...
+			getUserTable($_REQUEST);
+			break;
+		case 'updateUserPermissions':
+			updateUserPermissions($_REQUEST);
+			break;
+		case 'addUser':
+			addUser($_REQUEST);
+			break;
+		default:
+			die('Unregistered command. This instance has been reported.');
```
MatthewMeacham: lol good stuff

server/manageuser.php
```diff
@@ -0,0 +1,153 @@
...
+require_once 'config.php';
+require_once "user.class.php";
+
+$USER = new User();
+if (!$USER->hasPermission('edit_user_permission')) {
+	header("Location: /unauthorized");
+	die();
+}
+
+if(isset($_REQUEST['callback'])){
```
MatthewMeacham: On a sidenote, is this the preferred way of handling multiple request types through the same file? This 'callback' variable along with the request?

server/manageuser.php
```diff
@@ -0,0 +1,153 @@
...
+
+	$res = $stmt->execute(array(
+		$data['firstName'],
+		$data['lastName'],
+		$data['email'],
+		$data['phone'],
+		$data['prepareTaxes']
+	));
+
+	$response['success'] = !!$res;
```
MatthewMeacham: so is any value other than 0 a success? (And you're just transforming it to be either 0 or 1)

server/manageuser.php
```diff
@@ -0,0 +1,153 @@
...
+				(?, ?, ?)
+			ON DUPLICATE KEY 
+			UPDATE createdBy = ?");
+
+		foreach ($data['permissionIdArr'] as $permissionId) {
+			$stmt->execute(array(
+				$data['userId'],
+				$permissionId, 
+				$USER->getUserId(), 
+				$USER->getUserId()
```
MatthewMeacham: Duplicated here?

server/manageuser.php
```diff
@@ -0,0 +1,153 @@
...
+function updateUserPermissions($data){
+	GLOBAL $DB_CONN;
+	GLOBAL $USER;
+
+	$response = array();
+	$response['success'] = true;
+
+	$DB_CONN->beginTransaction();
+
+	$stmt = $DB_CONN->prepare("DELETE FROM UserPermission
```
MatthewMeacham: I'm not sure how I feel about how this method works. Correct me if I'm wrong, but it appears that you delete all the permissions they have in the database, and then add the ones that come in through $data. Is there any chance that we could change how we pass the data so we can see which ones need to be deleted and which ones need to be added? This way, we're not deleting all the permissions they have and then re-adding them. This way the audit trail of "createdBy" will be correct, and not just the last permission who edited the permissions.

Furthermore, I think this is a source of a bug. If a person removes/adds a permission to themselves, then it returns a 404 with an unauthorized. Then, the user can't navigate back to the /manage/user page.

MatthewMeacham:
It looks good other than that one bug where removing/adding a permission to yourself messes everything up.

manage/user/index.php
```diff
@@ -0,0 +1,89 @@
...
+	<title>Queue Test</title>
+	<?php require_once "$root/server/header.php" ?>
+	<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" />
+	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.12.4/css/bootstrap-select.min.css" />
+	<link rel="stylesheet" href="user.css" />
+</head>
+<body>
+	<?php require_once "$root/components/nav.php"; ?>
+	<div class="container">
+		<div class="row">
```
CollinsSpencer: Just a couple of styling things... The "row" starting on line 26 should probably have the "pt-5" class.

manage/user/index.php
```diff
@@ -0,0 +1,89 @@
...
+			<div class="col-md-9">
+				<h1>User Management</h1>
+			</div>
+			<div class="col-md-3">
+				<button class="btn btn-default pull-right" id="btn-add-user">Add User</button>
+			</div>
+		</div>
+		<div class="row">
```
CollinsSpencer: The "row" starting on line 34 should have its overflow-x set to auto. Currently the site kinda breaks on mobile, so this is an easy solution where people can just scroll horizontally without breaking the navbar. People probably won't be using this page on mobile, but it's a simple fix to keep our site from looking bad under most circumstances.

server/manageuser.php
2mnyzs: !! will cast 0/1 to false/true respectively. It's a shorthand thing basically.

server/manageuser.php
2mnyzs: Not sure on that, I've done it before as needed but it's usually taken care of by a framework in my experience.

manage/user/index.php
2mnyzs: Done

manage/user/index.php
2mnyzs: Done

server/manageuser.php
2mnyzs: I fixed the first part but it's still the source of a bug. Should I prevent users from removing their own permission that gives them permission to edit permissions?

server/manageuser.php
```diff
@@ -0,0 +1,157 @@
...
+
+	$response = array();
+	$response['success'] = true;
+
+	$DB_CONN->beginTransaction();
+
+	if(isset($data['removePermissionArr'])){
+		$stmt = $DB_CONN->prepare("DELETE FROM UserPermission WHERE userPermissionId = ?");
+
+		foreach ($data['removePermissionArr'] as $userPermissionId) {
```
MatthewMeacham: I'm not sure how I feel about this part. It would be better to do a bulk delete / insert rather than foreach-ing over all of them. But for right now we probably don't need to worry about it. 
The solution that comes to mind is basically adding the proper number of ? and then inserting an array. But I don't anticipate this permission changing handling extremely high volumes that it's necessary. So totally up to you

server/manageuser.php
```diff
@@ -102,6 +102,27 @@ function updateUserPermissions($data){
...
+		$stmt = $DB_CONN->prepare("SELECT lookupName 
+			FROM vita.permission
+				INNER JOIN vita.userpermission ON permission.permissionId = userpermission.permissionId
+			WHERE userPermissionId = ?");
+
+		foreach ($data['removePermissionArr'] as $userPermissionId) {
+			$stmt->execute(array($userPermissionId));
+			$lookupName = $stmt->fetch(PDO::FETCH_ASSOC)['lookupName'];
+
+			if($lookupName == 'edit_user_permission'){
```
MatthewMeacham: You shouldn't use the `==` operator for string equality, you should use the `===` operator. https://stackoverflow.com/a/8494892

MatthewMeacham: I suppose it's technically fine in this case since we know they both will be strings so there will be no type coercion, but just as a general good practice, I think `===` is better

MatthewMeacham: Also take a look at the answer below that one I posted, and then one below that. That second one down is especially interesting.

MatthewMeacham:
A few bugs:
* When adding a user, the prepares taxes checkbox doesn't actually change anything in the database (it is always 0 even if I click the checkbox)
* When adding a user, one can mash the mouse a bunch of times and add multiple of the same user to the database. Should probably disable the button on click, and idk maybe some JS thing to prevent multiple method requests from going through idk @CollinsSpencer

management/user/index.php
```diff
@@ -0,0 +1,89 @@
...
+	<div class="container">
+		<div class="row pt-5">
+			<div class="col-md-9">
+				<h1>User Management</h1>
+			</div>
+			<div class="col-md-3">
+				<button class="btn btn-default pull-right" id="btn-add-user">Add User</button>
+			</div>
+		</div>
+		<div class="row" style="overflow-x: auto;">
```
MatthewMeacham: We should probably avoid inline styles

management/user/user.js
```diff
@@ -0,0 +1,102 @@
...
+		$('#add-user-modal').modal({});
+
+		$('#add-user-form').on('submit', function(event){
+			event.preventDefault();
+			
+			var valid = true;
+
+			$(this).find('input').each(function(){
+				var hasVal = !$(this).val();
+				if(hasVal){
```
MatthewMeacham: This hasVal variable is kinda confusing. I would've expected it to represent if the input has a value, but you're using it to represent if the input doesn't have a value. Perhaps restructure it so you have:
```var hasVal = $(this).val();
if (!hasVal) {
...
}
```

server/manageuser.php
```diff
@@ -0,0 +1,178 @@
...
+
+	$thead = "<thead><tr><th>Name</th><th>Email</th><th>Permissions</th><th>Count</th></tr>";
+	$tbody = "<tbody>";
+	while($row = $stmt->fetch(PDO::FETCH_ASSOC)){
+		$optionList = getUserPermissionOptionList($row['userId']);
+
+		$tbody.= "<tr data-user-id='".$row['userId']."'>";
+		$tbody.= "<td>".$row['firstName']." ".$row['lastName']."</td>";
+		$tbody.= "<td>".$row['email']."</td>";
+		$tbody.= "<td><select class='userPermissionList selectpicker' multiple=true>".implode('', $optionList['options'])."</select></td>";
```
MatthewMeacham: I really don't like how you're specifying these CSS classes in this PHP code.. Do you think it would be better to just return the data back to the AJAX request, and then in the success function, create the table and append everything? Or why did you take this approach?

server/manageuser.php
```diff
@@ -0,0 +1,178 @@
...
+	$stmt = $DB_CONN->prepare("SELECT userId, firstName, lastName, email, preparesTaxes, archived 
+		FROM vita.user
+		ORDER BY firstName, lastName");
+
+	$stmt->execute(array());
+
+	$thead = "<thead><tr><th>Name</th><th>Email</th><th>Permissions</th><th>Count</th></tr>";
+	$tbody = "<tbody>";
+	while($row = $stmt->fetch(PDO::FETCH_ASSOC)){
+		$optionList = getUserPermissionOptionList($row['userId']);
```
MatthewMeacham: Also not sure how I feel about executing another SQL request for each user... Why not combine all of this into something (albeit a little uglier, but also better if you go the AJAX data / front-end table creation method, see comment below).
```
SELECT User.userId, firstName, lastName, email, preparesTaxes, User.archived, Permission.permissionId, Permission.name, userPermissionId 
FROM User
JOIN UserPermission ON User.userId = UserPermission.userId
JOIN Permission ON UserPermission.permissionId = Permission.permissionId
WHERE User.userId = ?
```

server/manageuser.php
```diff
@@ -0,0 +1,178 @@
...
+
+	$count = 0;
+	while($row = $stmt->fetch(PDO::FETCH_ASSOC)){
+		if($row['userPermissionId']){
+			$selected = 'selected=true';
+			$data = "data-userPermissionId='".$row['userPermissionId']."'";
+			$count++;
+		}else{
+			$selected = '';
+			$data = '';
```
MatthewMeacham: in the case where `$row['userPermissionId']` is not defined, do we really want to add something at all to this options array? Why not `continue` here? Or better yet, wrap it all in the `if($row['userPermissionId'])` block? I'm not understanding this else case.

server/manageuser.php
```diff
@@ -0,0 +1,178 @@
...
+	GLOBAL $DB_CONN;
+	GLOBAL $USER;
+
+
+	$response = array();
+	$response['success'] = true;
+
+	if($data['userId'] == $USER->getUserId()){
+		$stmt = $DB_CONN->prepare("SELECT lookupName 
+			FROM vita.permission
```
MatthewMeacham: Also is it good practice to specify the vita database schema in all this SQL queries? Should we be doing this all over?

server/manageuser.php
```diff
@@ -0,0 +1,178 @@
...
+			(firstName, lastName, email, phoneNumber, preparesTaxes)
+		VALUES 
+			(?, ?, ?, ?, ?)");
+
+	$res = $stmt->execute(array(
+		$data['firstName'],
+		$data['lastName'],
+		$data['email'],
+		$data['phone'],
+		$data['prepareTaxes']
```
MatthewMeacham: I think the issue with prepares taxes is right here! You forgot the "s" on "prepares". And also, do you think it's a good idea to transform the `$data['preparesTaxes']` to 0 or 1?

server/manageuser.php
2mnyzs: Not at all, I just have a bunch of databases locally and don't use defaults that often. I've removed them to be consistent with elsewhere.

server/manageuser.php
2mnyzs: Yes, the options where that is not defined are the possible permissions they don't have yet. Un-selected options so to speak. If we don't build them in, then the user can't add new permissions to someone because they(the ungranted permissions) won't be in the dropdown.

server/manageuser.php
2mnyzs: You would also need to select out a list of all permissions and then somehow workout the logic to attach unused permissions to each user. Or rather change how it's designed on the frontend. I wouldn't worry about performance on this page due to how often it will be used in practice.

server/manageuser.php
2mnyzs: I've done both(building HTML in JS and building HTML in PHP) enough to where I prefer PHP because it's more convenient. Template literals in JS are pretty nice if forced to build HTML client side. I build it inline in the HTML nowadays using VueJS, which I prefer even more. 
I can do client side if needed here.

management/user/user.js
2mnyzs: My bad, I originally had it meaning that it had a value, and then reversed it later on without switching the name. Should be better now like `noValue`.

management/user/index.php
2mnyzs: Changed to use an ID and matching selector.

server/manageuser.php
2mnyzs: It is better practice. Updated.

server/manageuser.php
2mnyzs: Adding a conversion client side fixed it. My bad on missing this. Thanks for catching it!

server/manageuser.php
MatthewMeacham: Let's not worry about doing it client-side for now. But my argument for client-side would be that we keep the front-end stuff more on the front-end rather than having front-end stuff in what is essentially our back-end. I would prefer more strict lines between layers, than letting them cross like this is doing. If one ever renamed the class, one would need to know that this is happening in the backend PHP which is kinda poor design in my opinion. 

But since this works, let's leave it for now and focus on more important things

server/manageuser.php
MatthewMeacham: Hm I think you're right about not worrying about this too much, but this is definitely a place to start optimizing if we ever needed to.

PR #106, opened 2017-11-02
Add Functionality to Log Out

CollinsSpencer:
I added functionality for logging out to the nav bar.

### DIFF
--- a/components/nav.php
+++ b/components/nav.php
@@ -30,7 +30,22 @@
 					<a class="nav-link" href="/queue">Queue</a>
 				</li>
 				<li class="nav-item">
+				<?php if ($USER->isLoggedIn()): ?>
+					<a class="nav-link" onclick="logout()">Log out</a>
+					<script type="text/javascript">
+						function logout() {
+							$.ajax({
+								url : "server/logout.php",
+								type: "POST",
+								success: function() {
+									location.reload();
+								}
+							});
+						}
+					</script>
+				<?php else: ?>
 					<a class="nav-link" href="/login">Volunteer Login</a>
+				<?php endif; ?>
 				</li>
 			<?php if ($USER->hasPermission('can_view_management_tab')): ?>
 				<li class="nav-item dropdown">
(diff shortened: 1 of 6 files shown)


server/logout.php
```diff
@@ -0,0 +1,7 @@
+<?php
+	$root = realpath($_SERVER["DOCUMENT_ROOT"]);
```
MatthewMeacham: Do you think this functionality would be better put in server/callbacks.php? That's where all the login/register/password reset stuff is, so I think it makes sense to keep all this functionality together in that file.

MatthewMeacham: on a sidenote, we should also rename server/callbacks.php I think...

2mnyzs: Agreed on both.

components/nav.php
```diff
@@ -30,7 +30,23 @@
 					<a class="nav-link" href="/queue">Queue</a>
 				</li>
 				<li class="nav-item">
+				<?php if ($USER->isLoggedIn()): ?>
+					<a class="nav-link" onclick="logout()">Log out</a>
+					<script type="text/javascript">
+						function logout() {
+							$.ajax({
+								url : "server/callbacks.php",
```
MatthewMeacham: This actually won't work on any page except the index

components/nav.php
```diff
@@ -30,7 +30,23 @@
 					<a class="nav-link" href="/queue">Queue</a>
 				</li>
 				<li class="nav-item">
+				<?php if ($USER->isLoggedIn()): ?>
```
MatthewMeacham: Also did we figure out why this PHP will fail on the Mac's MAMP?

2mnyzs:
Works great, LGTM.

components/nav.php
2mnyzs: Perhaps it used to be shorthand tags which was an issue for some?

MatthewMeacham: Update, got this fixed by rearranging some of the `require_once` in user.class.php and it ended up working 🤷‍♂️

PR #108, opened 2017-11-05
Redirect directories

MatthewMeacham:
The idea here is that many directories of ours lack an `index.php`, so one can navigate to the file path in the URL and see the folder structure and files in there. And since we don't control (to my knowledge) the apache settings on the server, this is a workaround that will create an `index.php` in every directory that doesn't have one, and that `index.php` will simply redirect the user back to the homepage. For example, check out http://cse.unl.edu/~vita/server/

I pulled the redirect code out into `server/redirect.php` so if instead we ever want to redirect instead to unauthorized or something, we can do that. 

I think that this should be put into our build process so that it is never the case that we forget to do this. I did not commit the files that this powershell script creates, as it is unnecessary in my opinion, but if y'all disagree we can commit those 100+ files; but I think the build process route is the way to go.

**Testing**
*  You can test this by simply running the powershell file.
*  After you do, make sure the directories all have a `index.php` file
*  Navigate to a directory that should redirect now such as `/server/libs`
*  Ensure that you are redirected to the index page
*  Clean your git repo
    *  run a `git clean -df`
    *  run a `cd /server/libs/PHPExcel`
    *  run a `git clean -df` again

### DIFF
--- a/scripts/CreateRedirectIndexPages.ps1
+++ b/scripts/CreateRedirectIndexPages.ps1
@@ -6,9 +6,10 @@ $path = '../'
 foreach ($directory in Get-ChildItem -recurse -dir $path) {
 	if (!(Test-Path ($directory.FullName + '\index.php'))) {
 		New-Item ($directory.FullName + '\index.php') -type file -value @"
-			"<?php 
-			$root = realpath($_SERVER["DOCUMENT_ROOT"]);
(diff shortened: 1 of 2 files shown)


2mnyzs:
Works as a good substitute for server configuration. Honestly, I'm not that concerned about people navigating to empty directories but this solves the issue regardless.

PR #109, opened 2017-11-05
Print Volunteer Schedules

MatthewMeacham:
Added in the functionality for admins to print the volunteer schedules for a site on a day.

I created a class that wraps the functionality for PHPExcel as this is what you should always do when interfacing with an API, but also just for ease of use; it removed a lot of the complexities of working with PHPExcel and hid it in one class. Additionally, as we create more Excel downloads, we can continue to use this and minimize duplicated code.

Also fixed a bug that would cause the PHPExcel to fail if the title of the sheet was longer than 31 characters (Excel itself only allows 31 characters in a sheet title)

**Pre-Testing**
* Run the test data powershell script
* Make sure you are not signed in on an account
* Try typing, directly in the URL bar: `localhost/management/documents`
    * It should show the unauthorized page
* Try typing, directly in the URL bar: `localhost/server/management/documents/volunteerSchedule.php`
    * It should show the unauthorized page
* Try typing, directly in the URL bar: `localhost/server/management/documents/appointmentsSchedule.php`
    * It should show the unauthorized page

**Testing**
* Login using siteadmin credentials
* Navigate to /management/documents using the dropdown in the nav ("Print Documents")
* Set the date to one month out from the current day (this is the first day for shifts lol)
* Hit Download Volunteer Schedule
* Verify it has two tabs, and data for volunteer schedules

**Additional Testing**
* Please verify that Download Appointment Schedule still functions correctly (Note you will need to change the date to the current date)


2mnyzs:
Tested, works great. +1

CollinsSpencer:
Props to you for getting the timezone stuff figured out. Looks good and seems to work fine.

Issue #71, opened 2017-08-19
As a developer, I want my JS files / CSS to be bundled and minified to reduce the number of files we have to send to a client.


MatthewMeacham:
It would be nice to have a sort of webpack type thing going on so that JS/CSS are bundled and minified when sent to the client in production. I don't know what sort of work this would take, but it would be a nice addition. Further, it would be nice to transcompile ES6 to ES5 so that our site will be wider supported. See http://caniuse.com/#search=ES6 and http://caniuse.com/#search=ES5

PR #110, opened 2017-11-08
Site, date, and time pickers

CollinsSpencer:
You should now be able to completely fill out the form. This includes selecting the date you want to make an appointment for, the site you want to make an appointment for on that date, and the time for the selected date/site.

Testing considerations:
- [x] you should not be able to select a site or time until a date is selected
- [x] you should not be able to select a site that is not available on the selected date
- [x] you should not be able to select a time until a date and site are selected
- [x] you should not be able to select a time that is not available on the selected date at the selected site
- [x] going back and changing an answer should not create any loopholes allowing you to do any of the above

Notes:
There is 1 TODO stub in signup.js. It has to do with being able to select the year that the appointment will be made during. It will be necessary to add additional functionality to load the shifts from a different year.

I also added two small external libraries to the project:
1. the jquery-ui.css for just the datepicker elements (since the js for jquery-ui is already being imported on this page, it doesn't need the js).
2. ~~the javascript and css for a jquery-timepicker.~~ We are no longer using this timepicker library


signup/index.php
```diff
@@ -314,12 +316,25 @@
...
 					</div>
 
-					<div id="appointmentPicker" style="height:300px">
-						Appointment Picker
-					</div>
+					<h3 class="form-subheading">Appointment Information</h3>
 
-					<div id="studentScholarAppointmentPicker" style="display:none; height:300px">
-						Student Scholar Appointment Picker
+					<div id="appointmentPicker" style="min-height:300px">
```
MatthewMeacham: Should I complain about inline styles?

signup/signup.js
```diff
@@ -13,6 +17,191 @@ $(document).ready(function() {
...
+		this.dates.push(date);
+		this.shiftTimes = [];
+		this.shiftTimes.push({
+			date: date,
+			startTime: startTime,
+			endTime: endTime
+		});
+	}
+
+	getDate(dateTime) {
```
MatthewMeacham: On a sidenote, private methods normally start with an underscore by convention: https://stackoverflow.com/a/8288762

signup/signup.js
```diff
@@ -13,6 +17,191 @@ $(document).ready(function() {
...
+			"endTime": true,
+			"siteId": true
+		}),
+		cache: false
+	})
+	.done(function(data) {
+		for(const shift of data) {
+			let sitesArray = sitesWithShifts.map(function(s) {return s.siteId});
+			if(sitesArray.includes(shift.siteId)) {
+				let i = sitesArray.indexOf(shift.siteId);
```
MatthewMeacham: I'm a little concerned about what is happening here. This method seems to assume that sitesWithShifts and sitesArray have matching indices for every siteId. I don't think this is an assumption we should make. You may be better off using a Map.

signup/signup.js
```diff
@@ -13,6 +17,191 @@ $(document).ready(function() {
...
+			"year": year,
+			"startTime": true,
+			"endTime": true,
+			"siteId": true
+		}),
+		cache: false
+	})
+	.done(function(data) {
+		for(const shift of data) {
+			let sitesArray = sitesWithShifts.map(function(s) {return s.siteId});
```
MatthewMeacham: Not that is matters, but why use a function(s) instead of the (s) => s.siteId ?

signup/signup.js
```diff
@@ -13,6 +17,191 @@ $(document).ready(function() {
...
+				}
+			});
+
+			if(availableTimes.length === 1) {
+				start = new Date(availableTimes[0].start);
+				end = new Date(availableTimes[0].end);
+				minTime = start.getHours() + ":" + start.getMinutes();
+				maxTime = end.getHours() + ":" + end.getMinutes();
+				$("#timeInput").timepicker('option', {'minTime': minTime, 'maxTime': maxTime});
+			} else if(availableTimes.length > 1) {
```
MatthewMeacham: What's the difference in logic between the ===1 case and the >1 case? They seem totally identical other than the sorting? And the sorting should work on an array of length 1, right? So couldn't these two cases be combined?

MatthewMeacham:
If one selects a site, and then a date, and then hits enter twice, it doesn't even require that one selects a time. And it selects like GMT time from 00:00 CST (so 06:00)

server/api/shifts/getAllShifts.php
```diff
@@ -0,0 +1,48 @@
...
+		}
+	}
+	$selectColumnsString = join(',', $selectColumns);
+
+	$year = 2018;
+	if(isset($data['year'])) {
+		$year = $data['year'];
+	}
+
+	$stmt = $DB_CONN->prepare('SELECT ' . $selectColumnsString . ' FROM Shift WHERE YEAR(startTime) = ' . $year);
```
MatthewMeacham: !!!!!!!
The year needs to be part of a prepared statement, not as string concatenation in order to avoid SQL injection.
I could pass in:
```
{
"year": "NULL OR 1==1"
}
```

In the current setup, this would return all the information. While this really isn't a big deal since shifts are all public anyways, it's good practice, and there's probably more malicious things one could do in the SQL that idk

signup/signup.js
```diff
@@ -13,6 +17,198 @@ $(document).ready(function() {
...
+	// 2. Update the options for the date selector to disable options that are not available at this site.
+	siteSelect.change(function() {
+		$("#datePicker").show();
+		$("#timePicker").hide();
+		// Clear any previously selected date
+		$("#dateInput").val('');
+		$("#timeInput").val('');
+		var value = this.value;
+		// Clear and repopulate the array of dates that are available at the selected site
+		datesAllowed.length = 0;
```
MatthewMeacham: On a sidenote, is this a valid way to empty an array? Or does this like basically trick JS into thinking the array is empty? What is happening here?

MatthewMeacham:
Oh uh also, did we agree that one should select a date first, followed by the site, followed by the time? Your current method does site->date->time which isn't what Vicky thinks we should do...

signup/signup.js
CollinsSpencer: Arrow functions are ES6, so the functionality would break if someone tried to use even IE11, which still comes standard with Windows 10 machines. Unfortunately, I also used the "class" syntax, which is also ES6, so it throws a syntax error if you open the page in IE11. We need to decide if we are supporting IE11 or not.

https://kangax.github.io/compat-table/es6/#ie11

signup/signup.js
CollinsSpencer: https://stackoverflow.com/a/1232046/7577035

signup/signup.js
MatthewMeacham: We are transpiling from ES6 to ES5 or even lower. I'm working on that ticket rn, #71

signup/signup.js
CollinsSpencer: Good catch

CollinsSpencer:
This PR should be GTG and is ready for review.

MatthewMeacham:
When I try to create a site, I get this in the console:
image

server/api/shifts/getAllShifts.php
```diff
@@ -0,0 +1,52 @@
...
+			if (isset($data[$key])) {
+				if ($data[$key] == true) {
+					$selectColumns[] = $key; // append the select columns
+				}
+			}
+		}
+	}
+	$selectColumnsString = join(',', $selectColumns);
+
+	$year = 2018;
```
MatthewMeacham: Really not sure how I feel about this. I think if for some reason the year isn't set, then we should outright reject the creation of the appointment. The most likely way that this would happen is if someone manually sent the AJAX request (which we don't wont), and if someone is doing that, then we probably don't want to make the appointment.

Imagine it is 2020, and someone creates an appointment without a year. Then we end up with an entry for an appointment back in 2018, which invalidates all the 2018 data. I would be partial to removing this and instead rejecting the creation if the year is not set.

server/api/shifts/getAllShifts.php
```diff
@@ -0,0 +1,52 @@
...
+	if(isset($data['year'])) {
+		$year = $data['year'];
+	}
+
+	$stmt = $DB_CONN->prepare('SELECT ' . $selectColumnsString . ' FROM Shift WHERE YEAR(startTime) = ?');
+	$stmt->execute(array($year));
+	$shifts = $stmt->fetchAll(PDO::FETCH_ASSOC);
+	foreach ($shifts as &$shift) {
```
MatthewMeacham: What's the point of the & in front of shift? Does this avoid you having to create a new array to pass back? Or what is happening here?

server/api/shifts/getAllShifts.php
CollinsSpencer: This is for retrieving the shift times in order to determine when to make appointments. I simply set it to have a default year that was valid. I think it would make more sense to have the default be ```$year = date("Y")```. The alternatives when an invalid year is sent in would be to either have no shifts be returned or have every shift be returned.

MatthewMeacham: Oh yah I forgot what file we were in. I think returning no shifts is the way to go with an invalid year

server/api/shifts/getAllShifts.php
CollinsSpencer: > "In order to be able to directly modify array elements within the [foreach] loop precede $value with &. In that case the value will be assigned by reference." [PHP Manual](http://pl.php.net/manual/en/control-structures.foreach.php)

MatthewMeacham: Good stuff, I didn't know that. Thanks Collins

server/api/shifts/getAllShifts.php
2mnyzs: > `$year = date("Y")`
> have every shift be returned

Either of the above are completely valid in my opinion.

signup/signup.js
```diff
@@ -57,7 +402,7 @@ function showAppointmentPicker(animationTime = 300) {
 }
 
 function showStudentScholarAppointmentPicker(animationTime = 300) {
-	$("#appointmentPicker").hide();
+	// $("#appointmentPicker").hide();
```
2mnyzs: Should avoid committing commented out code unless there's a reason for it.

PR #112, opened 2017-11-15
Transpiling JS

MatthewMeacham:
Aight this PR has Babel working for transpiling from ES6 to ES5. 

One does need NodeJS/NPM for all this to work:
* Download it here https://nodejs.org/en/
* Run this command in our project to install all the necessary packages `npm install`
* Now you can run the build.sh script

I spent a lot of time trying to get webpack to work successfully, but I couldn't get it to work, and I thought about it more, and I don't think we even need webpack for this at it's current stage. If y'all disagree, I can try again and reach out to some people I know who have used webpack before. 

This will place the the transpiled code in a /dist directory which is where any script inclusions will need to point. I changed them all in PR to point to the /dist. Additionally, what are the thoughts on checking the /dist directory into source control? Is this a good or bad idea? Should we ignore the /dist directory or not?

**Testing**
- [ ] Delete the /dist directory
- [ ] Run the build.sh script and make sure it runs successfully
- [ ] Ensure the /dist directory has code for every JS file
- [ ] Test every page and make sure the functionality is still correct


2mnyzs:
>Additionally, what are the thoughts on checking the /dist directory into source control? Is this a good or bad idea? Should we ignore the /dist directory or not?
This was the first thing I thought of. In most project's I've worked on we have ignored the /dist(and similar directories.) The idea being that git only contains actual source code. If you think there's a reason for keeping it in then I'm all ears.

PR #96, opened 2017-09-06
93 Database Changes

MatthewMeacham:
Aight went ahead and transitioned the Question / Possible Answer / Answer portion of the database to be ready for a hard-coded front-end question system. I added all of the litmus questions we know about right as well as all the branching questions we know of right now. 

Here is the new ERD:

databaseerd

PR #113, opened 2017-11-16
89 appointment management

andregarivay:
PR for the private queue. The queue now functions to progress clients through their appointment with buttons pressed by the greeter. Timestamps are sent to SQL each time the client moves forward in the appointment, and the appointment disappears after completion or a note on why it was left incomplete. There are a few buggy parts if you try to press the buttons too quickly, but that shouldn't be a problem, as the greeter will not be pushing a client through all four steps in a matter of seconds. Lots and lots of Angular... A few improvements can be made over time, but this version is fully functional.


components/nav.php
```diff
@@ -32,16 +32,6 @@
 				<li class="nav-item">
 					<a class="nav-link" href="/login">Volunteer Login</a>
 				</li>
-			<?php if ($USER->hasPermission('can_view_management_tab')): ?>
```
MatthewMeacham: @CollinsSpencer Did you ever get a chance to look into this on your Mac?

andregarivay: I wasn't able to ever figure out why Mac didn't like it. I did just update to High Sierra and I can check if compatibility with that changed for some reason.

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -73,9 +73,16 @@ CREATE TABLE Client (
...
+    scheduledTime DATETIME NOT NULL,
 	createdAt DATETIME NOT NULL DEFAULT NOW(),
-	arrivedAt DATETIME NULL,
+    language VARCHAR(255) NOT NULL,
+	timeIn DATETIME NULL DEFAULT NULL,
+    timeReturnedPapers DATETIME NULL DEFAULT NULL,
+    actualAppointmentTime DATETIME NULL DEFAULT NULL,
+    timeFinished DATETIME NULL DEFAULT NULL,
+    completed BOOLEAN NOT NULL DEFAULT FALSE,
+    notCompleted BOOLEAN NOT NULL DEFAULT FALSE,
```
MatthewMeacham: Hm I'm not sure how I feel about these `completed` and `notCompleted` columns. Because if completed is true, then notCompleted is obviously false and vice versa. The only edge case I see is that case where it's neither completed nor notCompleted (as in they haven't had the appointment yet), and I think NULL is a totally fine way to represent that situation. 
It's normally bad practice in a database to have fields that more or less depend on each other (like if we had a firstName and a lastName column, and then also a fullName column--it's redundant and kinda bad because then every time you update one field, you need to remember to change the other).
So I would be partial to removing either notCompleted or completed.

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -73,9 +73,16 @@ CREATE TABLE Client (
 
 CREATE TABLE Appointment (
 	appointmentId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
-	scheduledTime DATETIME NOT NULL,
+    scheduledTime DATETIME NOT NULL,
 	createdAt DATETIME NOT NULL DEFAULT NOW(),
-	arrivedAt DATETIME NULL,
+    language VARCHAR(255) NOT NULL,
```
MatthewMeacham: I think the plan for language was to have it as a Question asked on the signup form, and then leaving it as a question. But I think maybe it should actually be on Client. Idk we should talk this one through--I'm torn on where to put it

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -73,9 +73,16 @@ CREATE TABLE Client (
...
+	timeIn DATETIME NULL DEFAULT NULL,
+    timeReturnedPapers DATETIME NULL DEFAULT NULL,
+    actualAppointmentTime DATETIME NULL DEFAULT NULL,
+    timeFinished DATETIME NULL DEFAULT NULL,
```
MatthewMeacham: Just to verify, these column names are the ones that Vicky said they would want?

queue/queue.js
```diff
@@ -13,13 +13,27 @@ var queueApp = angular.module("queueApp", ["ngMaterial", "ngMessages"])
...
-				// TODO
 			} else if(data.length > 0) {
 				$scope.appointments = data.map((appointment) => {
 					// This map converts the MySQL Datatime into a Javascript Date object
 					var t = appointment.scheduledTime.split(/[- :]/);
 					appointment.scheduledTime = new Date(Date.UTC(t[0], t[1]-1, t[2], t[3], t[4], t[5]));
-
+					appointment.checkedIn = false;
+					if (appointment.timeIn != null) {
+						appointment.checkedIn = true;
```
MatthewMeacham: Couldn't you simplify these conditional statements to be of the form:
`appointment.checkedIn = appointment.timeIn != null;` 
And similarly for the ones below?

queue/queue_service.js
```diff
@@ -6,6 +6,76 @@ queueApp.factory("QueueService", function($http){
 			},function(error){
 				return null;
 			});
+		},
+		checkInNow: function(time, id) {
+			return $http({
+				url: "/server/queue_priv.php",
+				method: 'POST',
+				data: 'action=checkIn&time=' + time + '&id=' + id,
```
MatthewMeacham: You probably should change this to be in JSON form:
```
{
  action: 'checkIn',
  time = time,
  id = id
}
```
I believe it's the prefer way to specifying it like this, because this is harder to read and change

server/queue_priv.php
```diff
@@ -1,13 +1,83 @@
 <?php
```
MatthewMeacham: You should check to make sure a user is logged in before letting them continue on with calling this class.
Check out management/appointments/index.php. That does a permission check at the top, but you should do instead
`$USER->isLoggedIn()` and if the user is not, then don't let them call this page.

MatthewMeacham:
My browser is not liking the styling on this page. 
image

Other than those few comments above, this looks really good.

mysql_scripts/table_creation/FullDB.sql
andregarivay: Yeah, I looked back at my notes and this is what she had said she wanted. Essentially I just used the existing code with buttons instead of lanes.

queue/queue_service.js
CollinsSpencer: That's what I thought at first too. The problem here is actually that PHP can't handle Content-Type: application/json. We could write it as a JavaScript object, but then we'd have to override the Content-Type header and use a function to manually parse it into this same format. https://stackoverflow.com/a/20276775

andregarivay:
At this point, I'm having one small issue finishing up the styling, and the login verification was making it so I couldn't use the page (I can talk to someone about the verification later).

mysql_scripts/table_creation/FullDB.sql
andregarivay: This is something I'm hoping to cover in standup tonight

MatthewMeacham:
I imagine your issue with the login verification is that you didn't actually include the user.class.php 
```$root = $_SERVER["DOCUMENT_ROOT"];
require_once "$root/server/user.class.php"
```

components/nav.php
MatthewMeacham: Fixed in https://github.com/UNL-CSE-AMBASSADORS/VITA/pull/106

CollinsSpencer:
Styling fixes:

Add ```overflow-y: scroll;``` to ```.client-info-section```. It probably makes the most sense to include this inside the ```@media (min-width: 768px)``` section at the bottom of queue-private.css

Also, the styling on this page is a little weird in order to have sub-elements that are individually scrollable. You're setting the client-progress div to display as flex with the ```d-flex``` class, but that doesn't work in this situation because of browser reflow issues. Essentially, the children are being computed as only the size in which their internal text takes up (18px tall) and they are aligned inline. When the client-progress div is redrawn with its flex attributes, it doesn't push other elements out, which is what is causing all of the weird overlap. You can fix this by not using flexbox by removing ```d-flex``` and ```flex-column``` from the client-progress div. And then for each of the buttons add a class of ```my-1``` (small margin on top and bottom) and ```w-100``` (width of 100%).

mysql_scripts/table_creation/FullDB.sql
MatthewMeacham: I think that the `actualAppointmentTime` column should be renamed to `timeAppointmentStarted`. And perhaps rename `timeFinished` to `timeAppointmentEnded` so that we can be consistent between the start/end times? Also, I say ended because, to me, the word "finished" implies success, but ended doesn't necessarily imply success. It really doesn't matter though, but if you think that argument makes sense, feel free to change it. 

One other super important thing. How does one reconnect an appointment to what volunteer did that appointment? Your current structure doesn't allow that, right? At present, there is a table in the database called `ServicedAppointment` that was supposed to represent an appointment that a volunteer had at least started (not necessarily finished). I think you could use that table for these datetime columns you've added. And I think we can use that `ServicedAppointment` table to hold all this metadata-type information that goes along with an appointment. Additionally, that table already has a column for connecting it to the volunteer who did it (the userId column). You should also probably make that userId column nullable. We may want to rename the `ServicedAppointment` table, and feel free to if you think of a better name. 

Connecting who did an appointment with the appointment is extremely important for reporting purposes, and for correction purposes (if someone was doing something wrong, we have to know who all it effected). 

Again, consult https://github.com/UNL-CSE-AMBASSADORS/VITA/pull/96 to see the ERD and see what I mean.

queue/queue_service.js
```diff
@@ -6,6 +6,76 @@ queueApp.factory("QueueService", function($http){
...
+				return response.data;
+			},function(error){
+				return null;
+			});
+		},
+		turnInPapers: function(time, id) {
+			return $http({
+				url: "/server/queue_priv.php",
+				method: 'POST',
+				data: 'action=completePaperwork&time=' + time + '&id=' + id,
```
MatthewMeacham: On a sidenote, and there's definitely no need to change this. But we are going to be doing that transpiling from ES6 to ES5, so feel free to use ES6 features like string interpolation:
```
`action=completePaperwork&time=${time}&id=${id}`
````
That will do the same thing as the `+` signs, but it's more concise and tends to be easier to read and less verbose.

File: queue/queue_service.js [hunk 9]
```diff
@@ -1,7 +1,78 @@
...
+				return response.data;
+			},function(error){
+				return null;
+			});
+		},
+		incompleteAppointment: function(explanation, id) {
+			return $http({
+				url: "/server/queue_priv.php",
+				method: 'POST',
+				data: `action=appointmentIncomplete&explanation=${explanation}&id=${id}`,
```
MatthewMeacham: There is a pretty subtle bug here I believe. Imagine that within the explanation, I type the phrase "He was blah blah & blah blah". That & special character is actually going to screw up this URL, and your explanation won't be right in the backend. Also, imagine I did inside the explanation "&id=#", idk how it would handle the situation where there are two id things, but this is something we need to fix. See this post on using the encodeURIComponent: https://stackoverflow.com/a/15287088

server/queue.php
```diff
@@ -1,23 +1,33 @@
...
 	require_once "$root/server/config.php";
+	require_once "$root/server/user.class.php";
+	require_once "$root/server/utilities/dateTimezoneUtilities.php";
 	$conn = $DB_CONN;
 
 	// TODO make this handle multiple locations, if necessary
-	$stmt = $conn->prepare("SELECT DISTINCT appointmentId, scheduledTime, firstName, lastName
+	$stmt = $conn->prepare("SELECT Appointment.appointmentId, scheduledTime,
+		firstName, lastName, timeIn, timeReturnedPapers,
+		timeAppointmentStarted, timeAppointmentEnded, completed, notCompletedDescription
```
MatthewMeacham: Quick question on this stuff: Why is it necessary that for the queue we pass back `completed` and `notCompletedDescription`? The queue displays the upcoming appointments, I don't think we need to send this data to the front-end, especially for the queue. Furthermore, if an appointment is completed (or notCompleted), then it isn't even shown on the queue, right? I think it would be best to take these things out.

server/queue.php
```diff
@@ -1,23 +1,33 @@
...
 	$appointments = $stmt->fetchAll(PDO::FETCH_ASSOC);
 
 	// We must only display the first letter of the last name
 	// We do this server-side since we can't disclose the data client-side
-	for ($i = 0; $i < count($appointments); $i++) {
-		$appointments[$i]['lastName'] = substr($appointments[$i]['lastName'], 0, 1);
+	foreach ($appointments as &$appointment) {
+		$appointment['lastName'] = substr($appointment['lastName'], 0, 1);
+		$appointment['scheduledTime'] = getTimezoneDateFromUtc($appointment['scheduledTime'], $timezoneOffset)->format('Y-m-d H:i:s');
+		// TODO FINISH THE OTHERS
```
MatthewMeacham: You need to finish the others ones here (if they are not null), take a look at this article about figuring out if a variable is null: https://www.virendrachandak.com/techtalk/php-isset-vs-empty-vs-is_null/

server/queue_priv.php
```diff
@@ -3,12 +3,89 @@
 	require_once "$root/server/config.php";
 	$conn = $DB_CONN;
 
-	switch($_GET['action']) {
-		case 'display': displayAppointment($_GET['id']); break;
-		case 'cancel': cancelAppointment($_GET['id']); break;
+	// TODO: Make it so this is testable
```
MatthewMeacham: Remove this TODO, the code is testable

server/queue_priv.php
```diff
@@ -3,12 +3,89 @@
 	require_once "$root/server/config.php";
 	$conn = $DB_CONN;
```
MatthewMeacham: This is unused in the file, you can remove it

server/queue_priv.php
```diff
@@ -3,12 +3,89 @@
...
+	function completePaperwork($time, $id) {
+		$stmt = $GLOBALS['conn']->prepare(
+			"UPDATE ServicedAppointment
+			SET ServicedAppointment.timeReturnedPapers = ?
+			WHERE ServicedAppointment.appointmentId = ?"
+		);
+
+		$stmt->execute(array($time, $id));
+		$appointment = $stmt->fetchAll(PDO::FETCH_ASSOC);
+		echo json_encode($appointment);
```
MatthewMeacham: Does this actually return anything? I don't think UPDATE statements actually return anything, except maybe the number of entries updated? I'm fairly certain you could probably remove:
```
$appointment = $stmt->fetchAll(PDO::FETCH_ASSOC);
echo json_encode($appointment);
```

I'm not 100% confident, but I think so... I didn't see the result being used on the front end at least. I think the thing that we probably should be returning here is like a success/error, and if it errors then we should display a popup or something on the front-end saying it failed, and asking them to try again. It also provides that ackowledgement to the front-end so that the front-end can adjust the stylings if it wasn't successful. I'm not too worried about the error handling back to the front-end right now. But it is something we should do eventually. I'll spin off a ticket for this.

server/queue_priv.php
```diff
@@ -3,12 +3,89 @@
...
+	function appointmentStart($time, $id) {
+		$stmt = $GLOBALS['conn']->prepare(
+			"UPDATE ServicedAppointment
+			SET ServicedAppointment.timeAppointmentStarted = ?
+			WHERE ServicedAppointment.appointmentId = ?"
+		);
+
+		$stmt->execute(array($time, $id));
+		$appointment = $stmt->fetchAll(PDO::FETCH_ASSOC);
+		echo json_encode($appointment);
```
MatthewMeacham: Similar here about removing those lines, and similarly for all the UPDATEs in this file.

server/queue_priv.php
```diff
@@ -3,12 +3,89 @@
...
+		$stmt = null;
+	}
+
+	function appointmentComplete($time, $id) {
+		$stmt = $GLOBALS['conn']->prepare(
+			"UPDATE ServicedAppointment
+			SET ServicedAppointment.timeAppointmentEnded = ?, ServicedAppointment.completed = TRUE
+			WHERE ServicedAppointment.appointmentId = ?"
```
MatthewMeacham: Also just for the record, it's not necessary you change this, but just so you know. Whenever you only have 1 table, you don't actually need to prepend the table name whenever you're accessing a column. So this update could be expressed equivalently as:
```
UPDATE ServicedAppointment
SET timeAppointmentEnded = ?, completed = TRUE
WHERE appointmentId = ?
```

In fact, you never have to prepend a column with the table name unless multiple tables have the same column names

MatthewMeacham:
There is a small bug where you can click one of the buttons multiple times and it'll edit the database each time. Maybe this isn't a bug? Maybe it is? Idk how one could control it other than just not updating it if the column isn't NULL? 

However, what definitely is a bug is that if one hits the checkIn button multiple times then it creates multiple ServicedAppointment records for that appointment which gives us bad data, and will make the Queue display a name more than once.

Also, I don't like how I have to refresh the page in order to choose the next button if the queue hasn't updated yet. Or you have to click out of the appointment and back into it after the queue has updated. Is there a way to make that more responsive? Same thing with the "submit incomplete appointment" button.

Also, the greeter needs a select to choose what volunteer prepared the taxes for the person. And that will set the userId field in ServicedAppointment.

Also here is an example of that explanation bug that I was talking about: 
I forgot what I typed in, but it was basically like "&explanation="@!39482i12$@!(*#$(!(*#@%U%20%2021&id=62". You can see the result is clearly not what we intended.
image

server/queue_priv.php
andregarivay: When I remove those lines, the functions no longer work. I tried using just the server/config.php and that still didn't work.

MatthewMeacham: Sorry, I totally understand what is happening here now. So when you do this $conn = $DB_CONN, you're taking the global variable $DB_CONN and putting it into another global variable called $conn. And then you access this with that `$GLOBALS['conn']`. I think this is sort of redundant. I think what you should do instead is remove the `$conn = $DB_CONN`, and then do `$GLOBALS['DB_CONN']` instead. 

You will also see the structure in our codebase of 
```
GLOBAL $DB_CONN;
$DB_CONN->prepare(....);
```
Both the $GLOBALS associative array variable and GLOBAL keyword do the same thing (to my knowledge), so there is no difference, it's just a preference thing. Feel free to use whatever construct you like. 

But I do think you should not do this whole $conn = $DB_CONN thing. I can explain more about the PHP scoping in person when you have time.

server/queue.php
andregarivay: I took those out and realized that we still need to pull out the `completed` since the queue depends on that to decide whether or not to display an appointment. `notCompletedDescription` however, can be taken out.

MatthewMeacham: I mean should we even be passing back completed appointments at all? What's the point if it just hides it from the user anyway?

PR #116, opened 2017-11-21
78 qualification verification

MatthewMeacham:
Added in the ability for a site admin to verify the certifications/qualifications of a volunteer. I put this functionality on the permissions page that Schmidt had created before: /management/users

This PR basically copies what Schmidt had done for the permissions except it does the same thing for abilities instead. I also renamed some directories/files and move things around to improve consistency.

Also @2mnyzs why did you add in that "Permission Count" column? Is that just supposed to be a quick way to see who has a lot of permissions? 

# **To Test**
* Navigate to /management/users
* Add an ability to a person
    * Verify it is reflected in the database (the UserAbility table)
* Remove an ability from a person
    * Verify it is reflected in the database (the UserAbility table)
* Verify the permissions still works properly by removing/adding and verifying in databse


MatthewMeacham:
Pushed a change that standardizes those permissions and removed the unnecessary ones. Other PRs may be affected.

PR #128, opened 2017-11-30
Added in content and download EITC/VITA schedule ability for clients

MatthewMeacham:
Added in basic content, idk if this is actually good content, but I will get Virginia to proofread and suggest after this is deployed. The more important feature in this is the download file functionality. It can download files that are statically hosted on the box. This is used for clients to download the EITC schedule.


2mnyzs:
>  Virginia to proofread and suggest after this is deployed
Did she provide the content? If not, she probably should in my opinion.

MatthewMeacham:
She did not provide the content. And I agree, she _definitely_ should I had reached out to her a few weeks back about content, and I made the mistake of including "or should we just pull it from the other VITA sites". She said just pull it from other sites. However, I'll follow up with her on getting real content on this page. The purpose of this is to just get some generic stuff up there for the time being in case anyone reaches the website by mistake, and the more important update on this PR is the downloading the site schedule feature

PR #129, opened 2017-12-01
Unauthorized Page

JakeKo:
The unauthorized page now displays a 401 error message.


MatthewMeacham:
See text message about a possible change

PR #131, opened 2017-12-03
118 Production SQL Script

MatthewMeacham:
**UPDATE**: I want to discuss with @CollinsSpencer how DST will affect any of these times in the DB. 
**SO DO NOT MERGE**

Finished up the production data SQL script. This contains all the data that we need to manually insert for this year. 
Also changed the permission edit_user_permission to edit_user_permissions

**TO TEST**
Just run the FullDB.sql and then FullDBProdData.sql, and make sure there is no errors.


MatthewMeacham:
@CollinsSpencer We should actually talk about the data in here. I want to discuss how this will perform front-end once DST hits in March.

mysql_scripts/production_data/FullDBProdData.sql
```diff
@@ -0,0 +1,780 @@
...
+
+INSERT INTO Site (title, address, phoneNumber, appointmentOnly, createdBy, lastModifiedBy)
+	VALUES ("Bennett Martin Library", "14th and N Street", "402-472-9638", FALSE, @userId, @userId);
+SET @site_bennettMartinLibrary = LAST_INSERT_ID();
+
+INSERT INTO Site (title, address, phoneNumber, appointmentOnly, createdBy, lastModifiedBy)
+	VALUES ("Good Neighbor Center", "2617 Y Street", "402-472-9638", FALSE, @userId, @userId);
+SET @site_goodNeighborCenter = LAST_INSERT_ID();
+*/
+
```
2mnyzs: Please don't commit commented-out code.

2mnyzs:
I assume you're already aware but MySQL has a built in timezone function `convert_tz()`.

Issue #126, opened 2017-11-28
As a developer, I want the site to be styled in UNL template/branding.


MatthewMeacham:
See email exchange between Meach/Mike Jackson about this. But the site needs to be branded/templated with the UNL styling. This is an absolute must before it goes live January 15th. Focus on all public-facing parts of the site (administrative stuff can be changed over later if this is taking a while).

See this for what it should look like: https://lincolneitc.unl.edu/welcome

You should probably be able to just bring in the resources from https://wdn.unl.edu/ and change class names / add stuff / do whatever it is you need. For right now, just copy these resources into our project. A more sustainable solution will be thought through once this tax season is over.

CollinsSpencer:
https://www.unl.edu/wdn/templates_4.1/css/all.css

PR #135, opened 2017-12-04
124 dependent appointments

MatthewMeacham:
Created the idea of a DependentClient -- That is, our response to the situation where a client wants to sign up and bring their 4 kids to the appointment as well. So this enables a person to sign up themselves, and sign up their 4 dependents at the same time. What happens on the backend, is one appointment is created, one client is created, and then 4 DependentClients are created. 

Also sorry it doesn't look good @CollinsSpencer. Will let you restyle it when you do #126 ❤️ 

**TESTING**
* Go to the /signup page
* Hit the add dependent button -- Ensure that the two inputs / remove button gets added to the page
* Fill in all the information on the form and submit it
* Check the database and ensure that the DependentClient table has the correct information for what you entered
* Verify that the /management/documents download appointment schedule still works and contains the "Number of Returns" column.

**CHANGES TO PROD DB**
This script will need to be run against the PROD database prior to this being deployed in PROD. There is no impact of running this in prod before this PR is deployed.
```
CREATE TABLE DependentClient (
	dependentClientId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
	firstName VARCHAR(255) NOT NULL,
	lastName VARCHAR(255) NOT NULL,
	clientId INTEGER UNSIGNED NOT NULL,
	FOREIGN KEY(clientId) REFERENCES Client(clientId)
);
```


signup/index.php
```diff
@@ -60,6 +60,15 @@
 							<label class="form-label form-required" for="phone">Phone Number</label>
 						</div>
 
+
+						<!-- TODO CHANGE THE ADD DEPENDENT TO SOMETHING MORE DESCRIPTIVE OF WHAT THIS IS FOR CLIENTS-->
```
CollinsSpencer: Is this TODO to be done in this PR or is there another Issue set up to take care of it?

signup/signup.js
```diff
@@ -589,14 +601,28 @@ $('#vitaSignupForm').submit(function(e) {
 		}
 	});
 
+	var dependents = [];
+	$(".dependent-div").each(function() { 
+		var firstName = $(this).find('.firstName').val().trim();
+		var lastName = $(this).find('.lastName').val().trim();
+
+		if (firstName.length > 0 && lastName.length > 0) { // empty strings
```
CollinsSpencer: So if either the first or last name field is left empty, then it just simply won't add the dependent to the database and the client will have no way of knowing that the dependent they wanted to add did not get added.

I think it would be a better UX for those fields to both be required. If they try to submit the form without filling out one of them, it will throw an on screen error and allow them to go fix it. If they don't want to add the dependent, then they would just need to hit the remove button.

signup/index.php
MatthewMeacham: I think I would prefer to take care of it in this PR, however, the impact of doing it later is pretty minimal. All it involves in some simple renaming which is easy across the board (even in DB).

signup/signup.js
MatthewMeacham: Good call, will change now

MatthewMeacham:
@CollinsSpencer This is good for a second review, I changed up a few things to address your comments. And got confirmation from Moody/Vicky that "Dependents" is the right way to say this.

PR #140, opened 2017-12-08
Login process fixes

2mnyzs:
Fixed various issues with login process. The commits should be pretty self explanatory. Biggest fixes were the following:

- Fixed URLs in emails being relative
- Fixed register() not feeding into reset password when account is already registered
- Fixed reset password, had incorrect key reference

### DIFF
--- a/register/index.php
+++ b/register/index.php
@@ -21,8 +21,7 @@
 		<div class='container'>
 			<div class='row'>
 				<section class='col-sm-8 col-sm-offset-2 col-md-6 col-md-offset-3'>
-					<h1 class='text-center'>Hoovestol, inc.</h1>
-					<h1 class='text-center'>Eagle Express Lines, inc.</h1>
+					<h1 class='text-center'>VITA Lincoln</h1>
 					<hr />
 				</section>
 			</div>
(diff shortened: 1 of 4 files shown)


MatthewMeacham:
This looks good and seems to work. The only thing that I'm slightly concerned about is the now-generic error messages. I agree that in the event of an exception that we _didn't_ raise, that we should give a generic exception. But on errors that we _do_ raise, I think it's better if we give our descriptive error.
For example, in the case that the 'password' and the 'verifyPassword' don't match, it now says "Error reaching the server, try again later" when in reality, the only error is that the passwords don't match... A person would have no idea (bar JS checking of course, but that's not the point).

2mnyzs:
>  The only thing that I'm slightly concerned about is the now-generic error messages. I agree that in the event of an exception that we didn't raise, that we should give a generic exception. But on errors that we do raise, I think it's better if we give our descriptive error.
For example, in the case that the 'password' and the 'verifyPassword' don't match, it now says "Error reaching the server, try again later" when in reality, the only error is that the passwords don't match... A person would have no idea (bar JS checking of course, but that's not the point).

Good point, that was very sloppy of me. I added a private function that differentiates between the two types and passes back custom exceptions we raise. Thank you!

PR #144, opened 2017-12-15
136 move questions

MatthewMeacham:
Moved all the litmus questions to a /questionnaire page so that people can figure out if VITA can help them. We decided this would be best since some of those questions were confusing. 

Added in a link from the /signup page to the /questionnaire page.

**TESTING**
* Ensure that appointment sign up still works (Most importantly, the results are put in the Answers table correctly)
* Ensure that the questions on the /signup page still work as expected (like appearing/disappearing)
* Ensure that the questions on the /questionnaire page still work as expected (like appearing/disappearing)
* Ensure that answers that should make it so that VITA can't help them say that when one clicks that answer on the /questionnaire page


MatthewMeacham:
Yah so those last 5 questions don't have anything because that flier says that VITA "may be able to help" if they have/need any of those things. I wasn't sure what to do about that and figured we'd just leave them in. We can always remove them at Moody's/Vicky's request ¯\_(ツ)_/¯

PR #148, opened 2017-12-16
122 Client Information Permissions

MatthewMeacham:
This adds in a permission for a volunteer to 'view all client information' meaning specifically a client's last name, email address, and phone number. 

**TESTING**
* Run the database scripts
* Not logged in, go to the public queue and ensure that you only see a client's first name and last initial
* Logged into preparer@test.test, go to the private queue and ensure that you can see only a client's first name and last initial. Check the public queue, and it should be the same case. You should also not be able to see a client's email or phone number anywhere.
* Logged into siteadmin@test.test, go to the private queue and ensure that you can see a client's first name, full last name, phone number, and email (when you click on a client of course)

**Randomly adding in the new Database ERD here because the wiki won't let me upload images there, only link them, so I am creating a link in this PR**
databaseerd


server/queue_priv.php
```diff
@@ -97,29 +95,3 @@ function cancelledAppointment($id) {
 		echo json_encode($appointment);
 		$stmt = null;
 	}
-
```
MatthewMeacham: This all seemed like dead code from what I could tell, so I removed it.

queue/private.php
```diff
@@ -34,14 +34,14 @@
 
 				<div class="queue" ng-if="appointments.length > 0" ng-cloak>
 					<div class="row queue-row py-1 pointer"
-							 ng-repeat="appointment in appointments | orderBy:'scheduledTime' | searchFor: clientSearch"
+							 ng-repeat="appointment in appointments | orderBy:['timeIn == null', 'timeReturnedPapers == null', 'timeAppointmentStarted == null', 'scheduledTime'] | searchFor: clientSearch"
```
CollinsSpencer: I like this.

queue/private.php
```diff
@@ -111,7 +116,7 @@
 						 	</form>
```
CollinsSpencer: This form still has some bugs. For example, it doesn't clear when I switch between clients. Also, I can still type more than 255 characters into the field. The textarea has ng-maxlength="255" but it also needs maxlength="255". This issue doesn't match the scope of this ticket though. I just wanted to make note of it.

MatthewMeacham: Created https://github.com/UNL-CSE-AMBASSADORS/VITA/issues/152

PR #151, opened 2017-12-17
76 NEW Profile

MatthewMeacham:
Created the profile page for a volunteer. This is the place where they can edit their personal information, their abilities, see the certifications that have been verified by a site administrator, and also sign up for shifts.

There is a lot of moving parts here. Please read through the code carefully and test thoroughly.

**TESTING**
Test everything 😄 

* Run the database scripts
* Run the /scripts/build.sh script
* Log in as a user
* Navigate to the profile page by clicking the link in the nav bar
* Edit your personal information by hitting Edit.
  * Blank fields should result in a server error. Some JS magic might be nice here instead
  * A bad email should result in a server error.
  * Ensure result is saved properly in database
* Add an ability by clicking one in the multi select
  * Ensure result is saved properly in database
* Remove an ability by clicking the checked ability in the multi select
  * Ensure result is saved properly in database
* Manually give your userId the Basic Certification UserAbility with this command
```
INSERT INTO UserAbility (userId, abilityId, createdBy)
    VALUES ([INSERT_YOUR_USER_ID_HERE], 1, 5)
```
* Ensure that the X turns into a green check mark now in the Certifications section
  * Remove the ability and ensure it turns back to a red X
* Add a shift
  * If you get to the time select and then change the site, the other selects should reset themselves
  * After hitting Add, ensure that the result is properly saved in database
  * Ensure the cancel button works properly
  * Ensure the trash icon works properly (it should remove the shift from the page and database)
* Test anything else you can think of.


mysql_scripts/test_data/FullDBTestData.sql
```diff
@@ -146,11 +154,575 @@ SET @site_site2Id = LAST_INSERT_ID();
 INSERT INTO Site (title, address, phoneNumber, appointmentOnly, createdBy, lastModifiedBy)
 	VALUES ("No walkins site", "9876 Test St. Lincoln, NE 29384", "555-999-8888", true, @user_siteAdmin1Id, @user_siteAdmin1Id);
 SET @site_site3Id = LAST_INSERT_ID();
--- end site
 
+INSERT INTO Site (title, address, phoneNumber, appointmentOnly, createdBy, lastModifiedBy)
```
MatthewMeacham: This PR is significantly easier to test with all the data present with a bunch of data, we can remove this later. But it is nice to have while testing right now.

JakeKo:
Ran the aforementioned tests and tried breaking a few things on my own. Some notes:

* Remove the seconds from a shift's timestamp - it clutters up the dropdown
* Allow a person to cancel from changing profile settings

On the topic of email, I changed my email and logged out. Obviously, I couldn't log back in with the old email. Based on what I've seen in the wild, it seems there are two industry standards. Ask for a password confirmation and inform the person that this changes the email with which they login.

The other option, and my recommendation, is to send a confirmation email in conjunction with a warning that they are changing their email. Annoying to implement but is a small point that could save a serious headache or two.

Beyond that, I never encountered any actual problems or bugs. All field behaved as expected.

MatthewMeacham:
Decided that it would be best to disable email editing for right now. Created https://github.com/UNL-CSE-AMBASSADORS/VITA/issues/153 to address this.

profile/profile.js
```diff
@@ -0,0 +1,361 @@
...
+				let shift = response.shifts[i];
+
+				// Create the shiftsMap
+				if (!sitesMap.has(shift.siteId)) sitesMap.set(shift.siteId, shift.title);
+				if (!shiftsMap.has(shift.siteId)) shiftsMap.set(shift.siteId, new Map());
+				
+				let startDateTime = new Date(shift.startTime + ' CST');
+				let endDateTime = new Date(shift.endTime + ' CST');
+
+				let date = startDateTime.toLocaleDateString();
```
CollinsSpencer: This will not work if the user is in a different time zone. You can debug this by changing your system timezone in your Windows 10 Date & Time settings. You might have to restart your browser as well.

Instead, I would recommend formatting the shift time into a string on the server-side that is passed to the client as something like ```shiftTime: "5:30 PM - 8:00 PM"``` rather than send a startTime and endTime to the client to handle.

profile/profile.js
```diff
@@ -0,0 +1,361 @@
...
+				}
+			}
+		},
+		error: function(response) {
+			alert("Unable to load shift. Please refresh the page in a few minutes.");
+		}
+	});
+}
+
+let getTimeString = function(dateTime) {
```
CollinsSpencer: With this logic on the server-side, check out [this link](http://www.eltcalendar.com/stuff/datemysqlphp.html) for formatting. It's the most helpful list I've found.

profile/profile.js
```diff
@@ -0,0 +1,361 @@
...
+		dateSelect.change(function() {
+			timeSelect.children('option').remove();
+			timeSelect.append($('<option disabled selected value="" style="display:none"> -- Select a time -- </option>'));			
+			timeSelect.attr('disabled', false);
+			let siteId = siteSelect.val();
+			let dateString = $(this).val();
+
+			// populate shift times select now based on which date was chosen AND the user is not already signed up for it
+			for (const shift of shiftsMap.get(siteId).get(dateString)) {
+				if (shift.signedUp) continue;
```
CollinsSpencer: Rather than simply not showing the shift times that the user has signed up for, I would suggest disabling them. You can just add ```disabled: shift.signedUp ? true : false``` It will show the shift time but have it be disabled. I think this is more intuitive than simply having an empty drop-down.

profile/profile.js
```diff
@@ -0,0 +1,361 @@
...
+				cache: false,
+				success: function(response) {
+					if (response.success) {
+						shiftRow.remove();
+						// Find the shift and append the p tags to show the user is signed up
+						for (const shift of shiftsMap.get(siteId).get(dateString)) {
+							if (shift.shiftId === shiftId) {
+								let title = sitesMap.get(siteId);
+								appendSignedUpShift(title, dateString, shift.startTime, shift.endTime, response.userShiftId);
+								break;
```
CollinsSpencer: Since you aren't calling to the database again to re-pull all of the shifts. The shift that the user just signed up for is still available to be selected. If they try to sign up for the shift again, it just gets a generic error from the server.

The shift.signedUp value should probably be updated to true here to prevent the user from being able to try signing up for the same shift twice.

MatthewMeacham:
@CollinsSpencer Really good comments, it is significantly more responsive and intuitive now, and moving the conversion of time/date-to-string to the backend is way more flexible too. Thanks a ton on that.

rthakker549:
One thing that I see is that maybe we should disable the multi select unless the edit button is pressed because it is strange that you can edit the abilities. Another thing is that when I ran it last night, shift data would load, but now it doesn't load in leaving empty selects. You can see this here.
screen shot 2017-12-18 at 4 47 49 pm

MatthewMeacham:
@rthakker549 Yah I think the mutli-select is a little odd in this section at all. I had considered changing it to checkboxes instead, but never did. And I think I won't make the change simply because that stuff technically isn't even used anywhere yet.

And to your second comment, I believe the database schema changed, but I will verify.

Issue #130, opened 2017-12-02
As a greeter, I would like to know when my request didn't work on the queue so that I can try again or take the appropriate action.


MatthewMeacham:
Currently, the queue code doesn't give a success or failure when an action is made (like checkIn). It would be nice to send back a success/failure response to the user, and then display a popup or something asking them to try again if it didn't work, or ask them to contact the developers.

PR #154, opened 2017-12-18
37 queue site selection

andregarivay:
This PR is to allow the greeter to select the site from which they want to pull appointments.


queue/queue_header.php
```diff
@@ -1,9 +1,14 @@
+<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
+<script src="/queue/queue_header.js"></script>
 <div class="container-fluid dashboard bg-light py-3" ng-cloak>
 	<div class="d-flex flex-row justify-content-between align-items-center">
 		<div class="d-flex flex-row">
 			<div class="queue-size-lbl">Queue:</div>
 			<div class="queue-size-count">{{appointments.length}}</div>
 		</div>
+		<select class="btn dropdown-toggle" id="siteSelect" ng-model="selectSite">
+			<option value="">Select A Site</option>
```
CollinsSpencer: Currently this default option is selectable and shows up in the drop-down options. ```style="display:none"``` will make it not show up in the drop-down.
e.g. ```<option value="" style="display:none"> -- select a site -- </option>```

server/queue.php
```diff
@@ -8,7 +8,7 @@
 	// TODO make this handle multiple locations, if necessary
```
CollinsSpencer: You can remove this TODO now.

queue/queue_header.php
```diff
@@ -1,9 +1,14 @@
+<script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
+<script src="/queue/queue_header.js"></script>
 <div class="container-fluid dashboard bg-light py-3" ng-cloak>
 	<div class="d-flex flex-row justify-content-between align-items-center">
```
CollinsSpencer: I'm wondering if this should have ```d-flex flex-column flex-sm-row``` rather than ```d-flex flex-row``` so that the header elements stack on top of each other on small screens.

package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lockfileVersion": 1
```
CollinsSpencer: Make sure you're merging dev. package-lock.json should be in your .gitignore

CollinsSpencer:
Merging in dev will also help take care of the merge conflicts.

queue/queue.js
```diff
@@ -29,6 +34,20 @@ var queueApp = angular.module("queueApp", ["ngMaterial", "ngMessages"])
 			}
 		});
 	}
+
+	$scope.getSites = function() {
+		QueueService.getSites().then(function(data) {
+			if(data == null) {
+				console.log('server error');
```
2mnyzs: Might want to display an error to users when this happens. Correct me if I'm wrong but this would render the page inoperable.

queue/public.php
```diff
@@ -22,14 +22,12 @@
 	<!-- Body Section with list of clients -->
 	<div class="container-fluid queue" ng-if="appointments.length > 0" ng-cloak>
 		<div class="row queue-header py-2 bg-secondary text-light font-weight-bold">
-			<div class="col col-1 queue-id">Id.</div> <!-- TODO: Remove ID from here -->
-			<div class="col col-3 queue-name">Name</div>
+			<div class="col col-4 queue-name">Name</div>
 			<div class="col col-5 queue-progress">Progress</div>
 			<div class="col col-3 queue-time">Time</div>
 		</div>
 		<div class="row queue-row" ng-repeat="appointment in appointments | orderBy:'scheduledTime'" ng-if="appointment.completed == null" ng-class-even="'bg-light'">
```
2mnyzs: I would encourage breaking up lines like this with many different attributes, like you did below, it helps with readability.

queue/queue.js
MatthewMeacham: Yep, created https://github.com/UNL-CSE-AMBASSADORS/VITA/issues/130 to address this

PR #155, opened 2017-12-18
83 appointment scheduling rules

CollinsSpencer:
This PR removes logic about appointment times from the front-end and moves it all to the SQL queries and PHP. This also helped reduce the number of calls to the server from the client.

The database was updated to include an AppointmentTime table. This table is placed between Site and Appointment on the ERD. It takes holds individual datetimes for sites with the assumption that there is a Shift that covers the time frame of the AppointmentTime. By separating the times from the individual appointments, there will no longer be times placed in the appointment table that come from the front-end, i.e. all time stamps are set ahead of time, appointments just reference them. This also makes it easier to store logic surrounding the number of appointments that can be made. There are two columns that help with that:
* ```maximumNumberOfAppointments``` allows there to be a set number of appointments that can be scheduled. If not set, the number of appointments that can be made at a certain date/site/time can be calculated based on the number of volunteers that are signed up.
* ```percentageAppointments``` is used to limit the number of appointments that are made by the system, leaving spots for walk-in clients.

The test data was also updated to use predefined AppointmentTimes. In the test data Site 1 does not have a maximum, so appointment availability is based on the number of volunteers that have signed up. Site 2 has a maximum of 30 set. That being said, with 200 random appointments being made, many of the date/site/time options are full. Site 2 usually does still have availability.

**TESTING**

* Run the updated SQL scripts.
* Ensure that the API script works by visiting [localhost/server/api/appointments/getTimes.php](http://localhost/server/api/appointments/getTimes.php) (I recommend getting the Chrome Extension, JSONView to make reading this easier.)
* Go to the /signup page.
* Ensure that the date picker, site picker, and time picker all still work.
* Ensure that submitting the data still works and that it is stored correctly in the database.

**ERD**
Here's the ERD after these changes:
databaseerd


signup/signup.js
```diff
@@ -355,12 +116,13 @@ function updateSitesDatesAndTimes() {
...
+		dateSitesTimes.updateGlobalSites(date);
 		siteSelect.append($('<option disabled selected value="" style="display:none"> -- select an option -- </option>'));
-		for(const site of dateObj.getSitesArray()) {
+		for(let site in sites) {
 			siteSelect.append($('<option>', {
-				value: site.siteId,
-				text : site.title
+				value: sites[site]["siteId"],
+				text : sites[site]["title"],
+				disabled: sites[site]["hasTimeSlotsRemaining"] ? false : true
```
MatthewMeacham: You think we should disable these or just straight up not even show them?

signup/signup.js
```diff
@@ -372,16 +134,17 @@ function updateSitesDatesAndTimes() {
...
+		dateSitesTimes.updateGlobalTimes(date, site);
 		var timeSelect = $("#timePicker select")
 		timeSelect.append($('<option disabled selected value="" style="display:none"> -- select an option -- </option>'));
-		for(const time of siteObj.getTimesArray()) {
+		for(const time in times) {
 			timeSelect.append($('<option>', {
 				value: time,
-				text : time
+				text : time,
+				disabled: times[time] > 0 ? false : true
```
MatthewMeacham: Similarly here, should we disable it or just now show it? If we're disabling it, maybe we say something like "FULL" next to it so people know?

CollinsSpencer:
I made some of those small changes your pointed out. Thanks.

A thought that I had while making this ticket:

Currently, the total number of appointments is based on a ```maximumNumberOfAppointments``` column, if it is set, otherwise it is based on the number of volunteers that have signed up. However, it seems likely that they would want the capability of setting a ```minimumNumberOfAppointments``` value. Therefore, if not enough volunteers have signed up for a shift, a certain number of people can still make appointments, but if additional volunteers sign up for that shift, the number of possible appointments still increases with the number of volunteers.

```maximumNumberOfAppointments``` is intended for sites like the Anderson Library site or the International Student Scholar site that have a set number of appointments per appointmentTime.

Adding a ```minimumNumberOfAppointments``` column would not take away any functionality from the current system. Even if unused, the system would behave the same.

This would be fairly simple to add and does fit the scope of this ticket. Do you think I should add that functionality @MatthewMeacham ?

MatthewMeacham:
Go for it @CollinsSpencer

PR #159, opened 2017-12-21
Change from volunteer selection to station selection

MatthewMeacham:
After discussing with Moody/Vicky today, they thought it would be better if a greeter can select the station number instead of selecting the preparer that prepared the taxes. 

**UPDATES TO PROD DB**
The following script will have to be ran against the PROD database when this is deployed. There should be no impact to running this before, but I would nonetheless prefer waiting. 
```
ALTER TABLE ServicedAppointment DROP FOREIGN KEY ServicedAppointment_ibfk_1;
ALTER TABLE ServicedAppointment DROP COLUMN servicedBy;
ALTER TABLE ServicedAppointment ADD COLUMN servicedByStation INTEGER UNSIGNED NULL;
```

Note that I got the name of that foreign key constraint by running a `SHOW CREATE TABLE ServicedAppointment` which shows you the exact code one could use to recreate the table. Pretty neat.

### DIFF
--- a/mysql_scripts/table_creation/FullDB.sql
+++ b/mysql_scripts/table_creation/FullDB.sql
@@ -109,8 +109,7 @@ CREATE TABLE ServicedAppointment (
     timeAppointmentEnded DATETIME NULL DEFAULT NULL,
     completed BOOLEAN NULL DEFAULT NULL,
     notCompletedDescription VARCHAR(255) NULL DEFAULT NULL,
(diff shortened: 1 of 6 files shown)


MatthewMeacham:
@CollinsSpencer Yah I thought about making a database table for it, but it really didn't make much sense to me. Like it's always gonna numbers, and more than likely, always 1-25... So it doesn't really make sense to make a database call solely to get the values 1-25.

PR #165, opened 2017-12-22
Cache busting

MatthewMeacham:
Wired up the NPM production command to 'cache bust' our JS and CSS files. Meaning that it appends a unique hash (of the contents of the file) to the end of the file name, and then replaces every instance of the original file path with the cache-busted version.

This is obviously useful so that users of the site don't have to "hard reload" when using the site. The only thing that this doesn't do is remove the "old" transpiled files from the /dist directory, which really shouldn't be a big deal, the solution would be to create a new temp folder, but I didn't think it was worth it. Let me know if you disagree and I can add that in real quick.

**TESTING**
* Run `npm install` and `npm update`
* Run `npm run dev` (or the ./script/build.sh file) 
  * Ensure the website still works with no console errors / missing file errors / style errors
* Run `npm run production`, and you will notice that it replaces instances of our `/dist/**/*.js/css` files with `/dist/**/*-HASH.js/css`
  * Ensure the website still works with no console errors / missing file errors / style errors
  * Run `git reset HEAD --hard` to return to the state before the production build changes


package.json
```diff
@@ -4,8 +4,10 @@
   "description": "Scheduling and Appointment Management for VITA",
   "main": "index.js",
   "scripts": {
-    "dev": "babel . -d dist --ignore node_modules/,dist/",
-    "production": "babel . -d dist --ignore node_modules/,dist/"
+    "transpile": "babel . -d dist --ignore node_modules/,dist/",
+    "cachebust": "hashmark dist/**/*.js {dir}/{name}-{hash}{ext} | sed 's/\\\\\\\\\\\\\\\\/\\\//g'  | replaceinfiles -s **/*.php -d {dir}/{name}{ext}",
```
MatthewMeacham: I'm not certain why it required so many frigging escaped backslashes... But it seems that it actually ended up replacing escaped characters twice. Once when one runs the original command (`npm run production`), and a second time when the command was actually ran by npm... so I had to double escape everything I wanted. Not pretty, but it works.

signup/index.php
```diff
@@ -4,8 +4,8 @@
 <head>
 	<title>VITA Appointment Signup</title>
 	<?php require_once "$root/server/header.php" ?>
-	<link rel="stylesheet" href="/assets/css/form.css">
-	<link rel="stylesheet" href="/signup/signup.css">
+	<link rel="stylesheet" href="/dist/assets/css/form.css">
+	<link rel="stylesheet" href="/dist/signup/signup.css">
 	<link rel="stylesheet" href="/assets/css/jquery-ui-datepicker.css">
```
MatthewMeacham: I purposefully didn't prepend /dist to this `jquery-ui-datepicker` stuff since it's not ours, I see no reason to run it through our build process, I'm sure it's already been ran through a build process?

CollinsSpencer: That's totally fine. In fact, after the restyle, we won't need that file anymore since there is a UNL styled jquery ui datepicker.

JakeKo:
I'll post here about a problem @MatthewMeacham and I encountered off-stage:

Note that you may need to include './' if you're referencing a file in the same directory. The problem we encountered didn't appear on his machine but appeared on mine because his path in the file was `dist/` which my machine interpreted at `C:\dist\`. He fixed it with `./dist`.

Stay safe out there kids.

PR #169, opened 2017-12-23
167 return specifics

MatthewMeacham:
This enables a volunteer to select various statuses regarding the filing of a return. Namely, if it was e-file/paper for state/federal. This is something that was happening on the EITC Sign In form, and they wanted in the system. 

This system was created to be flexible enough to take on other types of things than just efile/paper for state/federal. Sometimes they amend taxes and sometimes even amend the amends. So in the future, if they want to keep track of that, we can easily add it in. The only limitation is that these things must be true/false type things.

**NOTE**: If anyone has a better name than FilingStatus and AppointmentFilingStatus for the two database tables, I'm all ears. I'm not sure if this is the proper wording or not for the flexibility that I want to portray.

**TESTING**
* Run RefreshDB script and build.sh script
* Log in, and go to private queue
* Click on a client, you should see the 4 checkboxes below the preparing button
  * These checkboxes should be disabled at the moment
* Click Checked In, Completed Paperwork, Preparing
* Click any of the checkboxes, and hit Finished
  * Ensure the ServicedAppointment entry has the right data for that client
  * Ensure that the AppointmentFilingStatus has the right data for the checkboxes you selected

**CHANGES TO PROD DB**
The following script will need to be ran against PROD when the changes are deployed.
```
CREATE TABLE FilingStatus (
    filingStatusId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
    text VARCHAR(255) NOT NULL,
    lookupName VARCHAR(255) NOT NULL,
    CONSTRAINT uniqueLookupName UNIQUE INDEX(lookupName)
);

CREATE TABLE AppointmentFilingStatus (
	appointmentFilingStatusId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
	servicedAppointmentId INTEGER UNSIGNED NOT NULL,
	FOREIGN KEY (servicedAppointmentId) REFERENCES ServicedAppointment(servicedAppointmentId),
	filingStatusId INTEGER UNSIGNED NOT NULL,
	FOREIGN KEY (filingStatusId) REFERENCES FilingStatus(filingStatusId)
);

INSERT INTO FilingStatus (text, lookupName)
	VALUES ('State E-file', 'state_efile');

INSERT INTO FilingStatus (text, lookupName)
	VALUES ('Federal E-file', 'federal_efile');

INSERT INTO FilingStatus (text, lookupName)
	VALUES ('State Paper', 'state_paper');

INSERT INTO FilingStatus (text, lookupName)
	VALUES ('Federal Paper', 'federal_paper');


```

### DIFF
--- a/queue/queue_service.js
+++ b/queue/queue_service.js
@@ -14,6 +14,13 @@ queueApp.factory("QueueService", function($http){
 				return null;
 			});
 		},
+		getFilingStatuses: function() {
+			return $http.get('/server/api/filingStatuses/getAll.php?filingStatusId=true&text=true').then(function(response) {
+				return response.data;
+			},function(error) {
+				return null;
+			});
+		},
 		getSites: function() {
 			return $http.get('/server/api/sites/getAll.php?siteId=true&title=true').then(function(response){
 				return response.data;
@@ -63,13 +70,16 @@ queueApp.factory("QueueService", function($http){
 				return null;
 			});
 		},
-		finishAppointment: function(time, id, servicedById) {
+		finishAppointment: function(time, id, servicedById, filingStatusIds) {
 			return $http({
 				url: "/server/queue/queue_priv.php",
 				method: 'POST',
-				data: `action=appointmentComplete&time=${time}&id=${id}&servicedById=${servicedById}`,
-				headers: {
-					'Content-Type': "application/x-www-form-urlencoded"
+				params: {
+					"action": "appointmentComplete",
+					"time": time,
+					"id": id,
+					"servicedById": servicedById,
+					"filingStatusIds[]": filingStatusIds
 				}
 			}).then(function(response){
 				return response.data;
--- a/mysql_scripts/production_data/FullDBProdData.sql
+++ b/mysql_scripts/production_data/FullDBProdData.sql
@@ -1748,6 +1748,24 @@ SET @permission_viewClientInformationId = LAST_INSERT_ID();
 
 
 
+-- FilingStatuses
+INSERT INTO FilingStatus (text, lookupName)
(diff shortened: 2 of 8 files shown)


queue/queue_service.js
```diff
@@ -63,13 +70,16 @@ queueApp.factory("QueueService", function($http){
...
 		},
-		finishAppointment: function(time, id, servicedById) {
+		finishAppointment: function(time, id, servicedById, filingStatusIds) {
 			return $http({
 				url: "/server/queue/queue_priv.php",
 				method: 'POST',
-				data: `action=appointmentComplete&time=${time}&id=${id}&servicedById=${servicedById}`,
-				headers: {
-					'Content-Type': "application/x-www-form-urlencoded"
+				params: {
```
MatthewMeacham: @CollinsSpencer Btw take a look at this. I ran into an error when I was trying to pass an array back through the url with form-encoded parameters. Did some googling, and it turns out there is this "params" parameter, that enables you to not have to do the form-encoded parameters through the url parameter.

CollinsSpencer: Yeah. You can't use the PHP's $_POST though because AngularJS $http defaults Content-Type to application/json, which $_POST doesn't parse. Since queue_priv.php uses $_REQUEST, which does parse application/json formatted parameters, the requests can be made with get and other types of requests in addition to just post requests.

Not a big deal here, but something to be aware of.

MatthewMeacham: ah I see, that is good to know.

CollinsSpencer:
This all looks good to me.

PR #170, opened 2017-12-23
168 appointment info

MatthewMeacham:
Added more information to the appointment confirmation page that a user sees after signing up successfully for an appointment. 

Also added in the ability for a client to print a printer-friendly version of that confirmation or email it to them if they entered an email. 

**TESTING**
* Run DB refresher, run build script
* Sign up for an appointment without an email.
  * You should get a lengthier appointment confirmation page now containing information on what to bring, etc. Please read through this and make sure there is no typoes.
  * At the bottom there should be a print button, click it and ensure the print dialog pops up
* Sign up for an appointment with an email
  * You should get the same appointment confirmation page
  * There should be a "Email Me this Confirmation" button at the bottom. Click it and the button should switch to "Sent!". Unless you have a mail server set up, you won't be able to test this functionality. But the network tab will have a response with the email confirmation message.


MatthewMeacham:
@CollinsSpencer I think the place to look up an appointment for a volunteer will be that page I am working on right now. It's probably going to become a more generic appointment management page. 

And 🤷‍♂️  on that other one. It is technically on the home page at the bottom, but not as in detail as the appt information/schedule.

CollinsSpencer:
This all looks good to me.

PR #171, opened 2017-12-24
Dependent/Language Information

MatthewMeacham:
Added in client language and dependents information to the private queue. 

**TESTING**
* Ensure the public queue still works
* Sign into Site admin account
* Go sign up for an appointment and add like 5 dependents
* Go to private queue, find the appointment you just created, ensure you see language and the dependents listed
  * Ensure you see the last name of the dependents
* Sign into preparer account
* Go to private queue, find the appointment you just created, ensure you see language and the dependents listed
  * Ensure you do not see the full last name of the dependents
* Ensure that appointments without dependents still show properly, and that the dependents heading is not there


queue/queue.js
```diff
@@ -26,7 +26,6 @@ var queueApp = angular.module("queueApp", ["ngMaterial", "ngMessages"])
 					appointment.paperworkComplete = appointment.timeReturnedPapers != null;
 					appointment.preparing = appointment.timeAppointmentStarted != null;
 					appointment.ended = appointment.timeAppointmentEnded != null;
-					appointment.name = appointment.firstName + " " + appointment.lastName;
```
MatthewMeacham: Whoops, this needs to stay here, this is used in the search functionality. My bad.

CollinsSpencer:
This all looks good.

I am curious though; with all of the dependents listed inside of the main appointment, what is the intended workflow for the progress/status? Since there are dependents, do those dependents not need to complete their own paperwork? If they are all just lumped together, then how it is currently is perfect. However, if they are helped individually, then it could be confusing.

MatthewMeacham:
@CollinsSpencer My understanding is that all dependents go and do one appointment together. So the current workflow system already encompasses this.

PR #172, opened 2017-12-24
130 private queue error

MatthewMeacham:
Added in some error handling around the private queue methods.

**TESTING**
* Run DB refresher and build script
* Ensure all private queue functionality still works properly
  * Observe that after one has clicked a private queue button, you can no longer click it again
* Force an error on one of the queue methods (Just go put a `throw new Exception()` in the try-catch of any of the methods in `/server/queue/queue_priv.php`
  * Ensure an alert pops up with an error message and that the $scope.client.xxxx is reset to false and the button is enabled again 
* Ensure the public queue still functions


queue/queue_private.js
```diff
@@ -31,19 +46,35 @@ queueApp.controller("QueuePrivateController", function($scope, $controller, Queu
 		}
 
 		$scope.client.ended = true;
-		QueueService.finishAppointment(new Date().toISOString(), $scope.client.appointmentId, $scope.client.selectedStationNumber);
+		QueueService.finishAppointment(new Date().toISOString(), $scope.client.appointmentId, $scope.client.selectedStationNumber).then(function(result) {
+			if(!result.success) {
+				$scope.client.ended = false;
+				alert(result.error);
+			}
```
CollinsSpencer: If the appointment is successfully marked as completed, it might just sit in the list of appointments until that's updated, which could be a full minute. I think we should call ```$scope.updateAppointmentInformation();``` if the result is a success. And same thing after being marked as incomplete or cancelled.

MatthewMeacham: Hm I hate the idea of hitting the database and loading all the appts just to get the change that they literally just made... I will however make this more responsive, let me see what I can do.

CollinsSpencer:
This all looks GTG.

PR #174, opened 2017-12-29
74 appt reschedule

MatthewMeacham:
Props to @CollinsSpencer for redoing this to fit the restyle and redoing the signup to use Angular.

This is technically done, but I'm not sure everything I did is "best practice" or "the better way to do things". @CollinsSpencer there's a few things I'm referring to in particular:

* Another thing... so the search functionality really only makes sense if it can search on _full_ first and last name... But if a volunteer doesn't have the 'view all client information' permission, then they could theoretically only search on first name and last initial, which isn't super intuitive and not very useful. What do we do about this? 
  * Do I just truncate the last name on the front-end for display and leave full last name for searching? The only problem is that this discloses the full last name client-side which is what we wanted to avoid in general, but idk.

**TESTING**
* Log in as a preparer, you should see a manage dropdown in the nav bar that only has appointments
* Log in as a site admin, you should see the manage dropdown in the nav bar that has appointments and the two others
* Navigate to /management/appointments
* Search for Barry, it should appear
* Click on the Barry appointment, it should go to a different page that has all the information and the ability to reschedule the appointment
* Select date/site/time, all that logic and stuff should work
* Hit reschedule, and the reschedule should be successful

**CHANGES TO PROD DB**
The following script needs to be run against the prod database when this is deployed:
```
ALTER TABLE Site
ADD COLUMN doesMultilingual BOOLEAN NOT NULL DEFAULT FALSE,
ADD COLUMN doesInternational BOOLEAN NOT NULL DEFAULT FALSE;

UPDATE TABLE Site
doesInternational = TRUE
WHERE siteId = 4;

UPDATE TABLE Site
doesMultilingual = TRUE
WHERE siteId = 3;
```


server/storeAppointment.php
```diff
@@ -90,7 +90,8 @@ function storeAppointment($data){
 				?,
 				(SELECT appointmentTimeId FROM AppointmentTime
 					WHERE DATE(scheduledTime) = ? 
-					AND time_format(TIME(scheduledTime), '%l:%i %p') = ?),
+					AND TIME_FORMAT(TIME(scheduledTime), '%l:%i %p') = ?
+					AND siteId = ?),
```
MatthewMeacham: @CollinsSpencer I also fixed what I think was a small bug here. I included the siteId in this query where you get the AppointmentTimeId for the time the user has inputted. Without the siteId, theoretically, if there was two AppointmentTimes that were happening on the same date at the same time, then this query would've returned 2 appointmentTimeIds and the whole query would've failed. Note that that is a valid case.

By including the siteId we should avoid that situation.

management/appointments/appointments.js
```diff
@@ -0,0 +1,115 @@
+let appointmentsApp = angular.module("appointmentsApp", ["ngMaterial", "ngMessages"]);
+
+appointmentsApp.controller("AppointmentsController", function($scope, $interval, AppointmentsService) {
```
CollinsSpencer: You don't need $interval here because it isn't used in the controller.

CollinsSpencer:
In general, using jQuery on top of AngularJS is not recommended. [Here's a quora post briefly answering it.](https://www.quora.com/Can-I-use-jQuery-with-AngularJS/answer/Maneesh-Tewani?srid=h3xQV)

I think it would be best to wait on this PR until after the restyle is merged. There are several changes that would need to be made anyways to get this to work with using RequireJS.

components/appointmentPicker/appointmentPickerController.js
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		$scope.$watch(
+			function() { return $scope.sharedProperties.studentScholar; }, 
+			function(newValue, oldValue) {
+				if(newValue) {
+					console.log("Student Scholar Appointment!");
+				} else {
+					console.log("Standard Appointment");
+				}
```
MatthewMeacham: Can probably remove these console.logs

signup/signup.php
```diff
@@ -0,0 +1,303 @@
...
+		<h3 class="form-subheading">Appointment Information</h3>
+
+		<div appointment-picker></div>
+
+		<div>Form valid? {{form.$valid}}</div>
+
+		<input type="submit" 
+			value="Submit" 
+			class="submit wdn-button wdn-button-triad" >
+			<!-- ng-click="storeAppointments()"> -->
```
MatthewMeacham: Remove commented out code?

signup/signupController.js
```diff
@@ -0,0 +1,100 @@
...
+
+			console.log($scope.data.firstName);
+			console.log($scope.data.lastName);
+			console.log($scope.data.email);
+			console.log($scope.data.phone);
+			console.log($scope.data.language)
+			console.log(questions);
+			console.log($scope.dependents);
+			console.log(scheduledTime);
+			console.log($scope.sharedProperties.selectedSite);
```
MatthewMeacham: Remove console.logs

signup/signupController.js
```diff
@@ -0,0 +1,100 @@
...
+				"email": $scope.data.email,
+				"phone": $scope.data.phone,
+				"language": $scope.data.language,
+				"questions": questions,
+				"dependents": $scope.dependents,
+				"scheduledTime": scheduledTime,
+				"siteId": $scope.sharedProperties.selectedSite
+			};
+
+			// $scope.successMessage = data;
```
MatthewMeacham: Remove commented out code

signup/signupController.js
```diff
@@ -0,0 +1,100 @@
...
+
+		$scope.addDependent = function() {
+			$scope.dependents.push({
+				firstName: '',
+				lastName: $scope.data.lastName ? $scope.data.lastName : ''
+			});
+		}
+
+		$scope.removeDependent = function(dependent) {
+			console.log(dependent);
```
MatthewMeacham: Remove console.log

signup/signupDataService.js
```diff
@@ -0,0 +1,33 @@
+define('signupDataService', [], function() {
+
+	function signupDataService($http, $httpParamSerializerJQLike) {
+		return {
+			storeAppointments: function(data){
+
+				console.log("post");
+				console.log(data);
```
MatthewMeacham: Remove console.log

signup/signup.php
```diff
@@ -0,0 +1,303 @@
...
+				</div>
+			</li>
+		</ul>
+
+
+		<h3 class="form-subheading">Appointment Information</h3>
+
+		<div appointment-picker></div>
+
+		<div>Form valid? {{form.$valid}}</div>
```
MatthewMeacham: Remove this

management/appointments/appointments.php
```diff
@@ -0,0 +1,70 @@
...
+		<!-- Show when there's no search results -->
+		<p class="wdn-inner-wrapper wdn-inner-padding-sm wdn-inner-padding-no-top wdn-center" 
+			ng-show="(appointments | searchFor: appointmentSearch).length == 0">
+			No results for "{{appointmentSearch}}"
+		</p>
+
+		<!-- List of appointments  -->
+		<table class="wdn_responsive_table queue" ng-show="(appointments | searchFor: appointmentSearch).length > 0">
+			<tbody>
+				<tr class="pointer"
```
MatthewMeacham: This class isn't taking effect on my machine

MatthewMeacham: Fixed btw

signup/signupController.js
```diff
@@ -37,13 +41,39 @@ define('signupController', [], function() {
 
 			SignupService.storeAppointments(data).then(function(response) {
 				if(typeof response !== 'undefined' && response && response.success){
-					$scope.successMessage = response.message;
+					document.body.scrollTop = document.documentElement.scrollTop = 0;
+					$scope.successMessage = $sce.trustAsHtml(response.message);
```
CollinsSpencer: This is cool.

MatthewMeacham: Yah if you don't trustAsHtml it actually just displays the code which is obviously not what we want, so to get around that, we have to do this.

signup/signup.php
```diff
@@ -0,0 +1,314 @@
...
+						<button type="button" class="wdn-button wdn-button-brand" ng-click="removeDependent(dependent)">Remove Dependent</button>
+					</div>
+				</ng-form>
+			</li>
+		</ul>
+		<br>
+		<button type="button" class="wdn-button wdn-button-triad" id="addDependentButton" ng-click="addDependent()">Add Dependent</button>
+
+		<style>
+		@media (min-width: 768px) {
```
CollinsSpencer: Should this get pulled out to the signup.css file?

MatthewMeacham: yes it probably should?

components/appointmentPicker/appointmentPicker.php
```diff
@@ -0,0 +1,78 @@
+<style>
```
MatthewMeacham: @CollinsSpencer Should these inline styles be moved to the appointmentPicker.css?

PR #179, opened 2018-01-14
123 shift cancel notification

MatthewMeacham:
This requires volunteers to specify why they are cancelling a shift they had previously signed up for. And then will send an email to the people specified in a .txt file in the repo.

**TESTING**
* Sign in as a volunteer
* Sign up for some shifts on the profile page
* Hit the cancel button, ensure that the cancellation reason modal pops up
  * If you don't specify a reason it shouldn't allow you to get past the Finish button
  * The X should exit the modal and not cancel the shift
  * The Close button should exit the modal and not cancel the shift
* Unless you have a mail server, you can't really test the email functionality. But you could instead test by adding in some code to `die($cancellationMessage);` (outside the `if(PROD)` block And review the result in the netwrok tab.


CollinsSpencer:
I think it would make sense to add the site/date/time/role info to the modal so that there isn't any confusion about which shift is being canceled.

There's also a couple functionality bugs. When you type a message and then hit close or the X, then select a different shift to cancel, and then finish, it will cancel both of those appointments. Additionally, if you go to cancel additional appointments, it will send an additional request back to the server for those shifts that you have previously canceled. Perhaps we need to move the ajax functionality to a more conditional state. Like the ajax only gets bound to the finish button in the modal once it opens up and then is removed from the button when the modal closes. Check out [http://www.jacklmoore.com/colorbox/](http://www.jacklmoore.com/colorbox/) in the Callbacks section. I don't know if this is the best solution, but it's the first thing that came to mind.

MatthewMeacham:
@CollinsSpencer Good catches, I'll make those changes tomorrow

PR #184, opened 2018-01-23
No-show appointments

MatthewMeacham:
After an appointment is 5 minutes late and still not checked in, it will now be marked as a "No-show" and displayed with a red pill on the public/private queues.

### DIFF
--- a/queue/private.php
+++ b/queue/private.php
@@ -44,7 +44,7 @@
 					ng-click="selectClient(appointment)">
 					<th class="queue-name" data-header="Name">{{appointment.firstName}} {{appointment.lastName}}</th>
 					<td class="queue-status" data-header="Progress">
-						<span class="pill pill-noshow" ng-if="appointment.noshow">No Show</span>
+						<span class="pill pill-noshow" ng-if="appointment.noshow">No-show</span>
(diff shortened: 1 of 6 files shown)


CollinsSpencer:
I like this.

In queuePrivateController.js, in the checkIn() function, should we add a line like ```$scope.client.noshow = false;```? You might have to store what the previous value was for the case of it not being a success though because we wouldn't want to default to saying that the user is a no-show if the server doesn't respond if they weren't a no-show before.

I'm not sure if this is necessary though. It gets corrected when you go back to the actual queue, so it would just be for the preparer who just checked the person in. Either we leave it like it is so that it shows the preparer or receptionist or whoever that the client WAS a no-show, or it will confuse the receptionist or whoever by thinking that they didn't actually get this person to not be marked as a no-show. I'm not sure what reaction they will have or which option is more intuitive.

MatthewMeacham:
I think we should add the `$scope.client.noshow = false`, that's a good call

MatthewMeacham:
@CollinsSpencer I was actually going to ask you about that as well. I'm not certain, it's easy enough to do either way, what do you think makes more sense?

PR #185, opened 2018-01-24
Update cache-bust:prod npm script

CollinsSpencer:
RequireJS references several js files but wasn't being updated to the newly cache-busted versions of files. This change should do the trick.


MatthewMeacham:
This is close, but doesn't work 100% properly. So in the appointmentPicker.js file when you specify the location of the appointmentPickerDataService, it's incorrect. You can see below that the /dist/components/appointmentPicker/appointmentPickerDataService was matched and replaced with the hashed version of that file, but /dist/components/appointmentPicker/appointmentPicker was also matched and replaced with the hashed version of that creating a path to a file that doesn't even exist.

appointmentPickerDataService: "/dist/components/appointmentPicker/appointmentPicker-8093568d960e7bfc367c273b36a319feafbcc03ee2c21196d5da55224ee1df0bDataService-260dfd1b0832893973b4179e909ed11c72005d76d84803d2a7527505981c92a0"

Is it possible to include the .js in these require paths? If so, that would solve our problems. Otherwise, we could change up the directory structure and that would fix this problem (but I would rather not do this). Other ideas?

MatthewMeacham:
Nicely done on this, works beautifully

PR #191, opened 2018-01-26
Create report to show appointment filing statuses

MatthewMeacham:
This creates a report that enables a volunteer to see the filing status of appointments that were serviced on a certain day.

**TESTING**
* Log in as siteadmin
* Go to the queue, and complete a few appointments
  * Select various filing statuses
  * Cancel an appointment
  * Incomplete an appointment with a reason
* Go to Print Documents, and hit the "Download Appointments FilingStatuses" button
  * Open the report. You should see a report that contains the appointment with whether or not it was completed, and the filing statuses.

### DIFF
--- a/server/management/documents/appointmentsFilingStatuses.php
+++ b/server/management/documents/appointmentsFilingStatuses.php
@@ -1,0 +1,144 @@
+<?php
+
+$root = realpath($_SERVER["DOCUMENT_ROOT"]);
+require_once "$root/server/user.class.php";
+$USER = new User();
+if (!$USER->hasPermission('use_admin_tools')) {
+	header("Location: /unauthorized");
+	die();
+}
+$HEADER_COLUMN_NAMES = array('First Name', 'Last Name', 'Appointment ID', 'Appointment Completed', 'Reason if Not Completed', 'State E-File', 'Federal E-File', 'State Paper', 'Federal Paper');
+$ALL_SITES_ID = -1;
+
+require_once "$root/server/config.php";
+require_once "$root/server/libs/wrappers/PHPExcelWrapper.class.php";
+
+getAppointmentsFilingStatusesExcelFile($_GET);
+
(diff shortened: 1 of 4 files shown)


server/management/documents/appointmentsFilingStatuses.php
```diff
@@ -0,0 +1,144 @@
...
+	if (empty($appointments)) {
+		$sheetIndex = $phpExcelWrapper->createSheet('None');
+		$phpExcelWrapper->setActiveSheetIndex($sheetIndex);
+		$phpExcelWrapper->insertHeaderRow($HEADER_COLUMN_NAMES);
+		$phpExcelWrapper->nextRow();
+	} else {
+		# Iterate through all the results and append them to the proper sheet
+		$sheetIndexForSiteId = array();
+		foreach ($appointments as $row) {
+			$siteId = $row['siteId'];
```
CollinsSpencer: This line doesn't exactly work since the query isn't pulling a siteId. As is, if you try to download the document for all sites, it will only create a sheet for the first site.

server/management/documents/appointmentsFilingStatuses.php
```diff
@@ -0,0 +1,144 @@
...
+
+	exit;
+}
+
+function executeAppointmentQuery($data) {
+	GLOBAL $DB_CONN, $ALL_SITES_ID;
+
+	$query = 'SELECT Client.firstName, Client.lastName, Appointment.appointmentId,
+		ServicedAppointment.completed, ServicedAppointment.notCompletedDescription, ServicedAppointment.servicedAppointmentId,
+		Site.title
```
CollinsSpencer: Change to ```Site.title, Site.siteId```

MatthewMeacham: Whoops, good catch, got a little too delete-happy when I was writing this query originally

PR #202, opened 2018-03-02
Fix overlapping shift time bug causing double appointments allotted during time

MatthewMeacham:
This PR attempts to fix the issue where during the 30-minute shift overlap, double the number of appointments are able to be scheduled (since it sees there are double the volunteers during that time). 

I'm not certain that this actually works. I'm having a hard time testing it on local test data or against prod SQL database. If nothing else, I don't think this will have an effect, but please use your vast SQL knowledge and determine if you think I've done this right.

**UPDATE**
I was able to test this against prod data and it functioned successfully. In the first one, I had made the naive mistake of assuming that subtracting two datetimes would yield the difference in minutes... idk what I was thinking. I'm not sure what the subtraction of two datetimes is, but it was something very large every time, so that old query was almost always true. This new one uses `TIMESTAMPDIFF(MINUTES, ...)` which _will_ return the difference in minutes and make this work properly. 
Also, I am open to changing the name of the field that I added, I wanted to capture that it was stored in minutes in the name, and also that it's not necessarily an exact appointment length time, and so `approximateLengthInMinutes` was the best I could come up with.

**CHANGES TO PROD DB**
The following SQL script needs to be run before this is deployed.
```
ALTER TABLE AppointmentTime
	ADD COLUMN approximateLengthInMinutes INTEGER UNSIGNED NOT NULL DEFAULT 60;
```

### DIFF
--- a/mysql_scripts/production_data/FullDBProdData.sql
+++ b/mysql_scripts/production_data/FullDBProdData.sql
@@ -142,24 +142,24 @@ INSERT INTO Shift (startTime, endTime, siteId, createdBy, lastModifiedBy)
 	VALUES (@shiftStartTime, @shiftEndTime, @site_andersonLibrary, @userId, @userId);
 
 SET @scheduledTime = "2018-01-24 16:30:00";
-INSERT INTO AppointmentTime (scheduledTime, minimumNumberOfAppointments, maximumNumberOfAppointments, siteId)
-	VALUES (@scheduledTime, 3, 3, @site_andersonLibrary);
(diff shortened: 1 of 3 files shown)


JakeKo:
I believe the query will work for all sites that have hour-long appointments. However, I believe it would prevent anyone from scheduling an appointment within the last 30 minutes of a shift.

JakeKo:
Good catch on that DateTime difference! Everything else appears to check out.

PR #223, opened 2018-03-16
205 reschedule email

MatthewMeacham:
Pulled out the HTML Email sending and Appointment Confirmation Message Generation into Utility class files with static methods. idk if this is best practice or the proper way to do it, so if you have any thoughts, I'm happy to hear them. 

I opted for the route that after an appointment is rescheduled, the system will automatically send an email to the client whose appointment was just rescheduled. I think this makes the most sense so that volunteers don't have to do anything. 

**TESTING**
Unless you have a local email server setup this is gonna be kinda hard to test
* Ensure after signing up for an appointment:
  * The appointment confirmation shows up
  * The "Email Me this Confirmation" button 'sends' the email (if you don't have an email server, you can check the network and see the response from the request, it should include the confirmation message--I think this is still in there)
* Go reschedule an appointment and make sure it still works
  * Not much you could do to test this, unless you got a local email server.


management/appointments/appointmentsController.js
```diff
@@ -91,6 +93,7 @@ define('appointmentsController', [], function() {
 
 		$scope.selectAppointment = function(appointment) {
 			$scope.appointment = appointment;
+			document.body.scrollTop = document.documentElement.scrollTop = 0;
```
MatthewMeacham: Just ease-of-use things that will make the screen jump back to the top when a volunteer clicks on a client or they reschedule an appointment on the appointment management page.

server/management/appointments/appointments.php
```diff
@@ -90,6 +99,28 @@ function rescheduleAppointment($appointmentId, $appointmentTimeId) {
...
+		$stmt = $DB_CONN->prepare($query);
+		$success = $stmt->execute(array($appointmentId));
+
+		if ($success == false) return;
+
+		$data = $stmt->fetch();
+		if (!isset($data) || !isset($data['email'])) return;
+
+		$confirmationMessage = AppointmentConfirmationUtilities::generateAppointmentConfirmation($appointmentId);
+		EmailUtilities::sendHtmlFormattedEmail($data['email'], 'VITA Appointment Rescheduled', $confirmationMessage, 'noreply@vita.unl.edu');
```
JakeKo: My gut says to put `noreply@vita.unl.edu` in a config file. Additionally, if `noreply@vita.unl.edu` is the default value in the method arguments, I see no reason to include it here explicitly.

MatthewMeacham: oooh good call, putting it in the config file is a great idea! And again, yah I'll remove that since it's the default

PR #224, opened 2018-03-16
Add more information to appointment management page

MatthewMeacham:
Added more information to an appointment on the appointment management page. 

**TESTING**
* Log in
* Go to queue page
  * Complete an appointment and record the client's name
  * Submit a cancelled appointment and record the client's name
  * Submit an incomplete appointment with a reason and record the client's name
  * Check in and paperwork complete an appointment and record the client's name
* Go to the appointment management page
  * Search for the completed client's name and click on that appointment
    * The status should be green and say "Complete"
    * All the time information on the right should be there (none should be N/A)
    * The Serviced at Station should be there
    * The Filing statuses you chose should be displayed
  * Search for the cancelled client's name and click on that appointment
    * The status should be red and say "Cancelled"
    * All the time information should be N/A
    * The Not Completed Description section should be there with "Cancelled Appointment" reason
  * Search for the incomplete client's name and click on that appointment
    * The status should be red and say "Incomplete"
    * The time information for how far you went should be there and the rest should be N/A
    * The Not Completed Description section should be there with the reason you gave
  * Search for the somewhat complete client's name and click on that appointment
    * The status should be blueish and say "In Progress"
    * The time information for how far you went should be there and the rest should be N/A
  * Click on any other appointment 
    * The status should be blueish and say "Not Started"
    * The time information should all be N/A

PR #225, opened 2018-03-16
203 cancel appointment

MatthewMeacham:
**DO NOT MERGE UNTIL #224 IS MERGED, THIS PR BUILDS OFF OF THAT ONE BECAUSE I NEEDED THE INFORMATION**

This PR enables a volunteer to cancel an appointment from the appointment management page. 

I pulled out the cancel appointment logic into a new server/accessors/appointmentAccessor.class.php class. Idk if accessor is the right word, or if doing this the class way is even the way to go. But I do want to share that code so that we don't have to duplicate it. Let me know if you have thoughts / other opinions on it.

**TESTING**
* Log in
* Go to appointment management page
* Click on an appointment that is in status "Not Started"
  * At the bottom, near the reschedule button, there should be a Cancel Appointment button
  * Click the button
  * It should be successful
    * The status text should now be "Cancelled"
    * The status text should now be red
    * The Cancel Appointment button should be gone
* Click on an appointment that is any status but "Not Started"
  * The Cancel Appointment button should not be there

### DIFF
--- a/management/appointments/appointmentsController.js
+++ b/management/appointments/appointmentsController.js
@@ -68,7 +68,7 @@ define('appointmentsController', [], function() {
 					$scope.submittingReschedule = false;
 					
 					// Let the user know it was successful
-					$scope.giveNotice("Success!", "This appointment was successfully rescheduled.", true);
+					$scope.giveNotice("Success!", "This appointment was successfully rescheduled.");
 				} else {
 					alert(result.error);
 
--- a/management/appointments/appointmentsController.js
+++ b/management/appointments/appointmentsController.js
@@ -4,6 +4,7 @@ define('appointmentsController', [], function() {
 
 		$scope.sharedProperties = sharedPropertiesService.getSharedProperties();
 		$scope.submittingReschedule = false;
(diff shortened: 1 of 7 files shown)


management/appointments/appointmentsController.js
```diff
@@ -53,39 +68,44 @@ define('appointmentsController', [], function() {
...
+				if (result.success) {
+					document.body.scrollTop = document.documentElement.scrollTop = 0;
+
+					$scope.appointment.cancelled = true;
+					$scope.appointment.notStarted = false;
+					$scope.appointment.statusText = "Cancelled";
+					$scope.appointment.notCompletedDescription = "Cancelled Appointment";
+
+					// Let the user know it was successful
+					$scope.giveNotice("Success!", "This appointment was successfully cancelled.", true);
```
JakeKo: The last argument is not necessary because you stipulate `true` is the default argument value.

MatthewMeacham: True, I always enjoy being more explicit, but the more I think about it, I think not specifying when I give a default is significantly better.

PR #233, opened 2018-05-23
Appointment Notes

MatthewMeacham:
This PR enables a volunteer to add notes to appointments and view those notes on the private queue or the appointment management page. Additionally, this **removes** the `notCompletedDescription` field from the `ServicedAppointment` table. This also updates the documents that an admin can download so that it no longer includes the notCompletedDescription data. This also adds a `cancelled` field to ServicedAppointment in place of the notCompletedDescription.

Also, @CollinsSpencer I would really like comments on how I display the notes. Not sure if just appending the author / timestamp to the note text is the correct approach. 

**TESTING ADDING NOTES**
1. Run the table creation and test data script
2. Log in as a preparer (no specific reason to be a preparer, but we under-test as a preparer, so just to make sure)
3. Navigate to the private queue
4. Select an appointment
    * You should see the note section which includes a textarea to add a note
5. Try to hit the "Add" button, you should not be able to
6. Type a note in the text area
7. Hit the "Add" button
    * The note should be added successfully
    * The note should appear in the area above and show that it was created by you a minute ago
8. Add another note
9. Copy the first name of the appointment so we can search it on the appointment management page
10. Choose another appointment on the private queue
    * You should not see the notes you put on the other appointment
    * Add a note to this appointment. Ensure it was successful
11. Navigate to the appointment management page
12. Search for the appointment you added notes to on the private queue
13. Ensure that you see the notes that you added
14. Add a note here. Ensure it was successful

**TESTING DOCUMENTS**
1. Please do this testing AFTER you did the above testing
2. Log in as the site admin
3. Go to the private queue and complete an appointment, be sure to mark any of the filing status options
4. Navigate to the "Print Documents" page
5. Select the current date and all sites. Download the "Appointment Filing Statuses" document
6. Open the Excel file
    * You should see the notes you added in the columns furthest to the fight
    * The filing statuses should be marked properly.

**MISCELLANEOUS TESTING**
I refactored some files and changed a lot of files due to removing that `notCompletedDescription` field. Could you please also test these functionalities:
* Rescheduling an appointment

**DEPLOYMENT STEPS**
The following steps will have to be taken once this is deployed to production.
1. Run the table creation SQL script
2. Run the cancelled updated SQL script
3. Run the /server/temp/DatabaseEvolution.php script **Please note that only Matthew will be able to run this**

**ENSURE THAT THE NOTES HAVE BEEN PROPERLY INSERTED AND THAT THE CANCELLED FIELD HAS BEEN SET CORRECTLY. ONLY PROCEED IF THAT HAS HAPPENED CORRECTLY**

4. Check again and make sure you are certain it is working
5. Run the notCompletedDescription update SQL Script 
6. Remove the /server/temp directory from prod and from master.

**TABLE CREATION SCRIPT FOR PROD**
```
CREATE TABLE Note (
    noteId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
    note VARCHAR(1000) NOT NULL,
    appointmentId INTEGER UNSIGNED NOT NULL,
    FOREIGN KEY(appointmentId) REFERENCES Appointment(appointmentId),
    createdBy INTEGER UNSIGNED NOT NULL,
    FOREIGN KEY(createdBy) REFERENCES User(userId)
);
```

**CANCELLED UPDATED SQL SCRIPT**
```
ALTER TABLE ServicedAppointment ADD COLUMN cancelled BOOLEAN NOT NULL DEFAULT FALSE;
```

**UPDATE SQL SCRIPT FOR PROD**
```
ALTER TABLE ServicedAppointment DROP COLUMN notCompletedDescription;
```

**NEW DATABASE ERD**
databaseerd

### DIFF
--- a/components/appointmentPicker/appointmentPickerController.js
+++ b/components/appointmentPicker/appointmentPickerController.js
@@ -5,7 +5,7 @@ define('appointmentPickerController', [], function() {
 		$scope.sites = [];
 		$scope.dates = [];
 		$scope.today = new Date();
-		$scope.sharedProperties = sharedPropertiesService.getSharedProperties();
+		$scope.appointmentPickerSharedProperties = sharedPropertiesService.getSharedProperties();
 
 		$scope.getAppointments = function(studentScholar = false) {
 			let year = new Date().getFullYear();
@@ -14,8 +14,8 @@ define('appointmentPickerController', [], function() {
 					alert('There was an error loading the appointments. Please try refreshing the page.');
 				} else {
 					$scope.dates = result.dates;
-					$scope.sharedProperties.hasAvailability = result.hasAvailability;
-					$scope.sharedProperties.isLoggedIn = result.isLoggedIn;
+					$scope.appointmentPickerSharedProperties.hasAvailability = result.hasAvailability;
+					$scope.appointmentPickerSharedProperties.isLoggedIn = result.isLoggedIn;
 					WDN.initializePlugin('jqueryui', [function () {
 						require(['jquery'], function($){
 							$("#dateInput").datepicker({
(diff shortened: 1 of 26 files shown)


components/appointmentPicker/appointmentPickerController.js
```diff
@@ -5,7 +5,7 @@ define('appointmentPickerController', [], function() {
 		$scope.sites = [];
 		$scope.dates = [];
 		$scope.today = new Date();
-		$scope.sharedProperties = sharedPropertiesService.getSharedProperties();
+		$scope.appointmentPickerSharedProperties = sharedPropertiesService.getSharedProperties();
```
MatthewMeacham: @CollinsSpencer I had to change the name of this because if they were both named `sharedProperties`, then they would overwrite each other. I don't fully understand scoping in Angular, so the easiest solution was to simply give this a more specific name that wouldn't conflict.

management/appointments/appointmentsController.js
```diff
@@ -21,11 +22,12 @@ define('appointmentsController', [], function() {
...
 							appointment.name = appointment.firstName + " " + appointment.lastName;
+
 							appointment.completed = appointment.completed == true; // Do this since the SQL returns 0/1, and we want it to be false/true
-							
-							appointment.cancelled = appointment.notCompletedDescription === "Cancelled Appointment";
+							appointment.cancelled = appointment.cancelled == true; // Do this since the SQL returns 0/1, and we want it to be false/true
 							appointment.notStarted = !appointment.cancelled && appointment.timeIn == null;
-							appointment.incomplete = !appointment.cancelled && appointment.notCompletedDescription != null;
+							// TODO: Not sure this logic for incomplete is good
+							appointment.incomplete = !appointment.cancelled && !appointment.completed;
```
MatthewMeacham: @CollinsSpencer Can you independently confirm this logic? I don't think it's actually sufficient, but I'm unsure of how to deem an appointment incomplete now that we don't have the notCompletedDescription

server/api/appointments/notes/notes.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
MatthewMeacham: I'd also like feedback on how I set this file up / where I put it. We haven't really been putting scripts like this in the api folder (and by "scripts like this", I mean a script that accepts an action and then delegates what to do based off of that parameter). So I'm not sure if this breaks our current architectural style, or if there would be a better place to put it.

MatthewMeacham: @CollinsSpencer

server/management/appointments/appointments.php
```diff
@@ -112,12 +102,18 @@ function rescheduleAppointment($appointmentId, $appointmentTimeId) {
...
-				completed = FALSE, notCompletedDescription = NULL, servicedByStation = NULL
+				completed = FALSE, servicedByStation = NULL
 			WHERE appointmentId = ?";
 		$stmt = $DB_CONN->prepare($query);
 
 		$success = $stmt->execute(array($appointmentId));
 		if ($success == false) throw new Exception();
+
+		// Add a note to the appointment saying it was rescheduled
+		$noteAccessor = new NoteAccessor();
```
MatthewMeacham: @CollinsSpencer Also thoughts on this automatic note addition? Yes or no?

CollinsSpencer: I think that is a really good idea!

CollinsSpencer:
FYI, all of the testing worked until I got to the miscellaneous testing section. I wasn't able to reschedule an appointment. The appointment picker didn't show up.

server/api/appointments/notes/notes.php
MatthewMeacham: Upon further thought, I think this is an ok place for it

MatthewMeacham:
**UPDATE**:
I forgot to update the `CREATE TABLE` SQL statement that was run to create the `Note` table to allow the `createdBy` field to be `NULL` (in which case the system assumes that it was created by SYSTEM). So I had to run the following query on the PROD database to fix my error.

```
ALTER TABLE Note 
    MODIFY COLUMN createdBy INTEGER UNSIGNED NULL;
```

PR #236, opened 2018-08-16
Add in a watch command for css/js

MatthewMeacham:
Should've done this a long time ago... especially considering the tools we were using already had this functionality built in! 

Anyway, here is an npm script to watch for changes in js/css files and run the `transpile-js` and `process-css:dev` scripts respectively.

**TESTING**
1. You will probably need to run `npm i` to install that npm-run-all dependency I added
2. Run `npm run watch`
3. Change a css file, observe as it re-processes that file for you 
4. Change a js file, observe as it re-transpiles that file for you
5. Rejoice cuz this will really speed things up, rather than running that `./scripts/build.sh` file all the time. I'm sorry I had y'all do that for so long.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,14 +5,17 @@
   "main": "index.js",
   "scripts": {
     "clean-start": "rm -rf ./dist/",
-    "transpile-js": "babel . -d ./dist --ignore ./node_modules/,./dist/",
-    "process-css:dev": "postcss ./**/*.css -d ./dist/ -b ./ --use postcss-cssnext",
-    "process-css:prod": "postcss ./**/*.css -d ./dist/ -b ./ --use postcss-cssnext cssnano",
+    "transpile-js": "babel . -d ./dist/ --ignore ./node_modules/,./dist/",
+    "process-css:dev": "postcss \"./!(dist|node_modules)/**/*.css\" -d ./dist/ -b ./ --use postcss-cssnext",
+    "process-css:prod": "postcss \"./!(dist|node_modules)/**/*.css\" -d ./dist/ -b ./ --use postcss-cssnext cssnano",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,14 +5,17 @@
...
     "minify-js:prod": "run-for-every-file --src \"./dist/\" --dest \"./dist/\" --file \"**/*.js\" --not-file \"**/*/\" --run \"uglifyjs {{src-file}} -o {{dest-file}} --compress drop_console --mangle --keep-fnames\"",
     "cache-bust:prod": "hashmark ./dist/**/*.{js,css} {dir}/{name}-{hash}{ext} | sed 's/\\\\\\\\\\\\\\\\/\\//g' > replace-map.json && cat replace-map.json | replaceinfiles -s ./**/*.php -d {dir}/{name}{ext} && cat replace-map.json | sed s/\\.js/\\\\\\x22/g | replaceinfiles -s ./dist/**/*.js -d {dir}/{name}{ext} && rm replace-map.json",
     "clean-end": "rm -rf ./dist/node_modules",
     "build:dev": "npm run clean-start && npm run transpile-js && npm run process-css:dev && npm run clean-end",
-    "build:prod": "npm run clean-start && npm run transpile-js && npm run minify-js:prod && npm run process-css:prod && npm run cache-bust:prod && npm run clean-end"
+    "build:prod": "npm run clean-start && npm run transpile-js && npm run minify-js:prod && npm run process-css:prod && npm run cache-bust:prod && npm run clean-end",
+    "watch-js": "npm run transpile-js -- --watch",
+    "watch-css": "npm run process-css:dev -- --watch",
```
MatthewMeacham: The `npm run` command enables you to pass additional arguments to the script you're calling by simply separating the command and the arguments with the `--` flag. Really nifty so that we didn't have to copy the `transpile-js`/`process-css:dev` commands into these watch commands, we can simply add the `--watch` argument through the `npm run` script!

PR #237, opened 2018-08-18
Self-Service Client Reschedule and Cancel

MatthewMeacham:
This PR enables clients to reschedule or cancel their own appointments without having to contact the center. I could've broken this feature into a few PRs, but oh wells ¯\_(ツ)_/¯

I did a lot of code cleanup in this PR since this feature led to some duplication between the appointment management functionality and this self-service functionality. So you will see quite a bit of that in here as well.

**PLEASE NOTE**: This PR pulls in both #233 and #236, so it should **not** be merged until those two are merged. It also may be hard to review in isolation until #233 is merged in.

**TESTING**
1. Run the DB creation script and the test data scripts
2. Navigate to the signup page and signup for an appointment
    * NOTE: To do this you will need to uncomment the comments in signup/signup.php that are preventing one from signing up. This is lines 4, 49, 88, 264. Note those line numbers are before removing any of the comment markers
    * ALSO NOTE: Remember all the personal information you put in! It is required in a later step
3. In the appointment confirmation, there is a new header for rescheduling and it provides a link.
4. Navigate to that link from step 3
    * NOTE: We prepend `https://` in the code, but your localhost probably isn't set up to handle that, so if the page doesn't load, you will probably need to remove the `https://` part
5. Welcome to the client reschedule and cancel page. 
6. Enter in information that *DOES NOT* match what you had put in when signing up.
    * It should say the information didn't match our records and clear all the input fields
7. Enter in information that *DOES* match what you had put in when signing up.
    * It should now show the reschedule/cancel functionalities
8. Scroll down and reschedule the appointment to any time, note which time you change it to
9. It should succeed, give a notification saying it succeeded, and then show the familiar appointment confirmation screen. 
    * Scroll down and the reschedule heading and information should also be there (note that the link is expected to be the same)
    * The print button should open the print functionality of the browser
    * The email me button should successfully send an email (you can see the message in the network tab) and change to say "Sent!" and be disabled after you hit it
10. Verify in the database that your appointment was rescheduled (alternatively verify on the appt management page)
    * Also, go to your appointment on the appt mgmt page, and there should be an automatic note saying the appointment was rescheduled by the client
11. Navigate again to your client reschedule link.
12. Enter in the correct information for your appointment
13. The current appointment information should display the new re-scheduled time.
14. Hit the "Cancel" button
    * A modal should popup asking you to confirm your cancellation
        * Clicking outside the modal box should close it with no cancellation performed
        * Clicking the X at the top right should close it with no cancellation performed
        * Clicking the "No, dont cancel" button should close it with no cancellation performed
15. Click the "Yes, Cancel" button
    * It should succeed, display a green success notification, and say your appointment has been cancelled
16. Verify the ServicedAppointment associated with your appointment is marked as cancelled and complete = false (either through the DB or appt mgmt page)
    * On the appt mgmt page for your appointment, there should be an automatic note saying it was cancelled by the client
17. Navigate again to your client reschedule link
18. It should say that this appointment can't be reschedule/cancelled and give the reason that it has already been cancelled
19. In the database, delete the ServicedAppointment entry that had been created for you
20. On the private queue, "Check In" your appointment
21. Navigate again to your client reschedule link
    * It should say can't be rescheduled/cancelled because the appointment has already been started
22. In the database, set the `timeIn` column on your ServicedAppointment entry to `NULL`
23. In the database, set the `scheduledTime` column on the AppointmentTime entry associated with your appointment to a date in the past
24. Navigate again to your client reschedule link
    * It should say can't be rescheduled/cancelled because the appointment's scheduled time is in the past
25. Change the token to something invalid (cases below) and it should say you've reached this page in mistake and tell them to click the link to go back home: 
    * More than 32 characters
    * Less than 32 characters
    * A token that doesn't have an associated appointment (just change one character in yours)

That is generally all the testing for this feature. You could also testing the "failure" notifications by adding `die()` calls in some of those private methods in the server/appointment/clientReschedule. I will not detail instructions for this however.

**MISCELLANEOUS TESTING**
Once again, I touched a lot of things making this PR, so a few things should be verified that they still work.
* Appointment Reschedule from the appt mgmt page
    * You should still get a success notification when the reschedule is successful
* Appointment cancellation from the appt mgmt page
* Cancellation from the private queue

**NEW DATABASE ERD**
databaseerd


appointment/reschedule/clientReschedule.js
```diff
@@ -0,0 +1,74 @@
+require.config({
```
MatthewMeacham: I wouldn't mind comments on the name of these files. You think `clientReschedule` is a good term? Or should we do something like `selfServiceReschedule`?

appointment/reschedule/index.php
```diff
@@ -0,0 +1,125 @@
...
+		</div>
+		<main id="wdn_content_wrapper" role="main" class="wdn-content-slide" tabindex="-1">
+			<div id="maincontent" class="wdn-main">
+				<div id="pagetitle">
+					<!-- TemplateBeginEditable name="pagetitle" -->
+					<h1>Cancel or Reschedule Appointment</h1>
+					<!-- TemplateEndEditable -->
+				</div>
+				<!-- TemplateBeginEditable name="maincontentarea" -->
+				<div id="clientRescheduleApp" class="wdn-band" client-reschedule token="<?php echo $token; ?>"></div>
```
MatthewMeacham: This is the way I'm passing the token value to the client-reschedule app, not sure if this is the best approach, but I couldn't just get the `$_REQUEST` in the appointments.php, I could only get it in the index.php for some reason, so I opted to pass the token as a parameter to our angular module. @CollinsSpencer thoughts?

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -4,6 +4,8 @@ DROP TABLE IF EXISTS Answer;
 DROP TABLE IF EXISTS AppointmentFilingStatus;
 DROP TABLE IF EXISTS FilingStatus;
 DROP TABLE IF EXISTS ServicedAppointment;
+DROP TABLE IF EXISTS AppointmentClientReschedule;
```
MatthewMeacham: Not sure if this is a good name for this table... thoughts? Maybe something like AppointmentRescheduleToken? This table holds the token and appointmentId right now. And in the future maybe failedValidationCount/lockoutTime if we deem it necessary to have security / throttling around thee self-service reschedule stuff. @CollinsSpencer @JakeKo

server/accessors/appointmentAccessor.class.php
```diff
@@ -11,13 +11,65 @@ class AppointmentAccessor {
 	public function cancelAppointment($appointmentId) {
 		GLOBAL $DB_CONN;
 
-		$query = "INSERT INTO ServicedAppointment (appointmentId, notCompletedDescription, completed)
-			VALUES (?, 'Cancelled Appointment', FALSE)";
-		$stmt = $DB_CONN->prepare($query);
+		$query = 'INSERT INTO ServicedAppointment (servicedAppointmentId, appointmentId, cancelled, completed)
+			VALUES (
+				(SELECT sa2.servicedAppointmentId FROM ServicedAppointment sa2 WHERE appointmentId = ?),
```
MatthewMeacham: So in this new query, we don't assume that a ServicedAppointment has not been created for the appointment. We use this `INSERT INTO _ ON DUPLICATE KEY UPDATE` clause that MySQL supports which inserts the entry if there exists no entry with the PK, and if the PK exists, it updates the entry based on what comes after the UPDATE portion of the clause. So the inner query here finds the PK of the entry if one exists, otherwise it returns NULL, and conveniently, if you pass NULL as the PK in MySQL, it just does the typical auto_increment functionality. Kinda neat.

For reference: https://dev.mysql.com/doc/refman/5.5/en/insert-on-duplicate.html

server/appointment/reschedule/clientReschedule.php
```diff
@@ -0,0 +1,267 @@
...
+	echo json_encode($response);
+}
+
+/* 
+ * Private functions
+ */
+
+function isAppointmentValidForReschedule($appointmentInformation) {
+	// See if it has been cancelled
+	$isAppointmentCancelled = isset($appointmentInformation['cancelled']) && $appointmentInformation['cancelled'] == true;
```
MatthewMeacham: Just wanna throw this out because it is something that I continually run into and forget about shortly thereafter... But MySQL actually returns 0/1 for false/true respectively, and so we can't use `===` to compare the result field value and a boolean literal because then we're comparing an integer to a boolean which is false in the `===` world. So one can explicitly cast to the same data type and use `===` or alternatively use `==`, but I think I prefer explicitly casting and then using `===`.

mysql_scripts/table_creation/FullDB.sql
CollinsSpencer: I don't have a problem with the `AppointmentClientReschedule` name, but I think `AppointmentRescheduleToken` might better fit the naming pattern with nouns for table names.

CollinsSpencer:
Overall, this PR looks great. I want to dig into the code a little more before giving it the go-ahead, but pretty much everything throughout the testing looked really nice.

I did find one thing that I think is worth changing. I think the security check on the Cancel or Reschedule page is slightly too strict. The email is being checked as case-sensitive, but emails are inherently case-insensitive. A similar thing is happening for phone numbers where it will fail if they are formatted differently e.g. (XXX) XXX-XXXX vs XXX-XXX-XXXX. If there is any way we can make the comparisons only look at the lowercase/uppercase version of names and emails and then strip phone numbers down to just the digits, then I think that will be better.

appointment/reschedule/clientReschedule.js
MatthewMeacham: @CollinsSpencer I decided to go with `selfServiceAppointmentReschedule` instead of `clientReschedule`--I think this new name describes what it is much better.

PR #239, opened 2018-08-18
Cancel Confirmation Dialog for Volunteers

MatthewMeacham:
Added in a cancel confirmation dialog for volunteers when they cancel an appointment. This code is similar to what is happening for the self-service client rescheduling.

**TESTING**
1. Sign in as a preparer
2. Go to appt mgmt page
3. Select any appointment
4. Scroll down and hit the cancel button
    * It should open up a dialog prompting the user to confirm they want to cancel the appointment
        * Clicking outside the dialog should close it and not cancel the appt
        * Clicking the X at the top right of the dialog should close it and not cancel the appt
        * Clicking the "No" button should close the dialog and not cancel the appt
5. Hit "Yes"
    * The appointment should now show as cancelled, the dialog should automatically close, and a success notification should appear


management/appointments/appointmentsController.js
```diff
@@ -139,6 +139,24 @@ define('appointmentsController', [], function() {
 			$scope.appointment = null;
 		}
 
+		$scope.initializeCancelConfirmationModal = function() {
+			WDN.initializePlugin('modal', [ function() {
+				require(['jquery'], function($) {
```
jjstobbe: The UNL Colorbox example does it like this:
```
    WDN.initializePlugin('modal', [function() {
        var $ = require('jquery');
        $('.group1').colorbox({rel: 'group1'});
    }]);
```
It's a little tough to read with like 4 functions in a row here.

Also, can you guys use arrow functions? That would be fantastic here to clean it up a little.

MatthewMeacham: Good note on the UNL Colorbox example--it works properly and is much cleaner. Changed to that. Additionally, I will start using arrow functions over regular `function` declarations.

PR #241, opened 2018-08-18
Use EmailUtilities

MatthewMeacham:
Use the EmailUtilities utility class in all the places the`mail()` function was being used.

**TESTING**
Unless you have a mail server set up locally, you probably won't be able to test this in it's entirety. However, this one is fairly simple so hopefully it's not an issue
1. Ensure you can still log in
2. Ensure the password forgot/register still works
3. Ensure the cancelling a shift still works

### DIFF
--- a/server/Login.class.php
+++ b/server/Login.class.php
@@ -298,7 +298,7 @@ class Login
 			if (!PROD) {
 				print $mail_body;
 			}
-			$this->sendHtmlFormattedEmail($dbemail, $subject, $mail_body, $this->noreply_email);
+			$this->sendHtmlFormattedEmail($dbemail, $subject, $mail_body);
 
 			$response['success'] = true;
 		}catch(Exception $e){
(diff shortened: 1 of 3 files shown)


server/Login.class.php
```diff
@@ -294,13 +294,11 @@ public function register($email){
 				$mail_body .= "<br />You can reach us at: <a href='mailto:".$this->contact_email."'>".$this->contact_email."</a>";
 			}
 
-			## Build Email Headers
-			$headers = "From: ".$this->noreply_email."\r\n";
-			$headers .= 'MIME-Version: 1.0' . "\r\n";
-			$headers .= 'Content-type: text/html; charset=iso-8859-1' . "\r\n";
```
jjstobbe: Loving the code clean-up. It's pretty dumb to be copy-pasting this stuff around.

server/Login.class.php
```diff
@@ -685,4 +677,12 @@ private function processException($exception, &$response){
 		}
 		$response['success'] = false;
 	}
+
+	private function sendHtmlFormattedEmail($toEmail, $subject, $body, $fromEmail) {
```
jjstobbe: Is it really necessary to have a `fromEmail` parameter if this application will only ever use the noreply email? Even the `sendHtmlFormattedEmail` function defaults to the noreply.

I would be on-board to get rid of `fromEmail` entirely, but maybe there are use-cases for it I'm not aware of 🤷‍♂️

MatthewMeacham: It this class, probably not. Good idea, removed that parameter and made it a local variable to the function.

PR #243, opened 2018-08-20
Limit volunteers on shifts

MatthewMeacham:
## Overview
This PR enables the functionality of limiting the number of volunteers who can sign up for a certain role at a site/shift. 

## Testing
1. Run the database scripts
2. Log in as preparer
3. Navigate to profile page
4. Scroll down to the shift sign up section
5. Hit the "Add a shift" button
    * Select "Test Site" -> 1st date option -> only time option
        * When you go to select a role, it should say there is a 10 person limit on the preparer role
    * Sign up for that shift in the preparer role
6. Hit the "Add a shift" button
    * Select "Test Site" -> 2nd date option -> only time option
        * When you go to select a role, it should say there is a 15 person limit on the preparer role
    * Sign up for that shift in the preparer role
7. Hit the "Add a shift" button
    * All the selects except the site select should be disabled
    * Once you select a site, the date select should be enabled, and the other two should be disabled
    * Once you select a date, the time select should be enabled, and the role select should be disabled
    * If you change the site/date/time the selects to the right should reset (and become disabled if applicable)
8. Log out and sign in as siteadmin
9. Navigate to the profile page
10. Hit the "Add a shift" button
    * Select "Test Site" -> 1st date option -> only time option
        * It should say there is a 10 person limit on the preparer role, and it should say there are 9 spots remaining
    * Select the 2nd date option -> only time option
        * It should say there is a 15 person limit on the preparer role, and it should say there are 13 spots remaining

## Production database evolution
The following script will need to be run against the production database when this feature is deployed.
```
CREATE TABLE RoleLimit (
    roleLimitId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
    maximumNumber INTEGER UNSIGNED NOT NULL,
    roleId INTEGER UNSIGNED NOT NULL,
    FOREIGN KEY(roleId) REFERENCES Role(roleId),
    siteId INTEGER UNSIGNED NOT NULL,
    FOREIGN KEY(siteId) REFERENCES Site(siteId),
    shiftId INTEGER UNSIGNED NULL,
    FOREIGN KEY(shiftId) REFERENCES Shift(shiftId),
    CONSTRAINT uniqueRoleSiteShiftTuple UNIQUE INDEX(roleId, siteId, shiftId)
);
```

## New Database ERD
databaseerd


profile/profile.js
```diff
@@ -98,12 +104,67 @@ WDN.initializePlugin('modal', [function() {
...
+		// Maps a roleId -> shiftId -> maximumNumber of that role allowed for that shift
+		const shiftRoleLimitsMap = new Map();
+
+		function loadRoleLimits() {
+			$.ajax({
+				url: '/server/api/roles/limits/getAll.php',
+				type: 'GET',
+				dataType: 'JSON',
+				cache: false,
+				success: function(response) {
```
JakeKo: The `success` and `error` methods are deprecated and are removed in jQuery 1.8. I would recommend using `AJAX.done` and `AJAX.error` if you want to upgrade jQuery in the future. I don't believe there is a performance difference.

profile/profile.js
```diff
@@ -149,68 +224,68 @@ WDN.initializePlugin('modal', [function() {
...
-				inline:true,
-				width:'50%',
-				onOpen: function() {
-					$('#cancellation-reason-details').text(`${siteTitle}: ${dateString} ${startTimeString} - ${endTimeString} (${roleName})`);
-					
-					// modal submit functionality
-					$('#cancellation-reason-form').on('submit', function(event) {
-						event.preventDefault();
-						$('#cancellation-reason-form button[type=submit]').prop('disabled', true);
+				.text(' [Cancel]')
```
JakeKo: This is quite unfortunately deep. Consider refactoring this.

profile/profile.js
```diff
@@ -301,13 +368,61 @@ WDN.initializePlugin('modal', [function() {
...
+			});
+	
+			$('#abilitiesSelect').on('change', function(event){
+				// abilities to be removed - all non-selected options with set ids
+				let removeUserAbilityIds = $(this).children('.editView').children('input:not(:checked)[data-userAbilityId]').map(function(index, ele){
+					return ele.dataset.userabilityid; // note this is case-sensitive and should be all lowercase
+				}).get();
+	
+				// abilities to be added - all selected options w/o ids
+				let addAbilityIds = $(this).children('.editView').children('input:checked:not([data-userAbilityId])').map(function(index, ele){
```
jjstobbe: I'm not confident this is doing what you think it's doing. If this is the normal javascript `.map` on an array, then the first argument is the element, and the second argument is the index (you have it switched).

You might be trying to just do:
```
blahblah.map((ele) => ele.value)
```
(same problem with the `.map` above this one)

profile/profile.js
jjstobbe: Going off this, have you considered making generic Ajax methods that abstract some of this boiler-plate out of the way? Usually they're helpful for generic error tracking and if you run into issues like this where parts of jquery is depreciated, you'd only have to change it in one place.

profile/profile.js
MatthewMeacham: Wow, this has been here since literally as long as I can remember. Amazing catch.

profile/profile.js
```diff
@@ -98,12 +104,67 @@ WDN.initializePlugin('modal', [function() {
...
+				error: function(response) {
+					alert('Unable to load role limits. Please refresh the page in a few minutes.');
+				}
+			});
+		}
+
+		function addToSiteRoleLimitsMap(siteRoleLimit) {
+			const hasRoleIdAlready = siteRoleLimitsMap.has(siteRoleLimit.roleId);
+			if (!hasRoleIdAlready) {
+				siteRoleLimitsMap.set(siteRoleLimit.roleId, new Map());
```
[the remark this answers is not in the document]
MatthewMeacham: Yah, if you saw the page it might make a little more sense. But the idea is when people sign up for a shift they have to select a site, date, time, and role. The times are determined by which date they select, and the dates are determined by which site they select. And now the role is determined by which site/date/time they select.

profile/profile.js
MatthewMeacham: It's never crossed my mind, but we should definitely be doing this...

profile/profile.js
```diff
@@ -149,68 +224,68 @@ WDN.initializePlugin('modal', [function() {
...
-										$.colorbox.close();
-										
-										// Find the shift that just got removed and make it so we're not signed up for it anymore
-										for (const shift of shiftsMap.get(siteId).get(dateString)) {
-											if (shift.userShiftId === userShiftId) {
-												shift.signedUp = false;
-												shift.userShiftId = null;
-												break;
+							if(valid) {
+								$.ajax({
```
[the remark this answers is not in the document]
MatthewMeacham: lmao this is hilarious--and I definitely did write this, although not in this PR. Will look into refactoring this

profile/profile.js
MatthewMeacham: Ah, after doing some testing, this is actually not a JS `.map` this is a JQuery `.map` which has argument list `(index, element)`
See here: http://api.jquery.com/map/

jjstobbe: I wasn't sure if it was the normal `.map`, but nice job looking into it. Just gives me another reason to hate JQuery.

PR #246, opened 2018-08-24
Mark Appointments as Walk-Ins

MatthewMeacham:
This PR adds a "Walk-In" pill to the queue when an appointment is a walk-in. The way we determine if an appointment is a walk-in is if the createdAt time is > scheduledTime.

**TESTING**
1. Update an Appointment to have a createdAt time greater than it's scheduledTime. You can use the following query to identify appointments, their scheduled time, and their site title for the "Test Site" site.
```
USE vita;
SELECT a.appointmentId, at.scheduledTime, s.title
    FROM Appointment a
    JOIN AppointmentTime at ON a.appointmentTimeId = at.appointmentTimeId
    JOIN Site s ON at.siteId = s.siteId
    WHERE s.siteId = 1
    ORDER BY scheduledTime;
```
The following query can be used to update an appointment then to make it a walk-in appointment:
```
USE vita;
UPDATE Appointment
    SET createdAt = 'SOME_TIME_PAST_THE_SCHEDULED_TIME'
    WHERE appointmentId = INSERT_APPOINTMENT_ID_HERE
```
2. Navigate to the public queue and select the "Test Site" and proper date
3. The appointment you changed should now have a green "Walk-In" pill next to it now, and it should be the very last appointment shown for the scheduledTime
4. Log in as preparer
5. Navigate to the private queue and select the "Test Site" and proper date
6. The appointment you changed should now have a green "Walk-In" pill next to it now, and it should be the very last appointment shown for the scheduledTime
7. Check in any appointment and the appointment that is a Walk-In
8. The Walk-In appointment should be the last one displayed for the Checked-In + ScheduledTime pair

### DIFF
--- a/queue/private.php
+++ b/queue/private.php
@@ -39,13 +39,13 @@
 			</thead>
 			<tbody>
 				<tr class="pointer"
-					ng-repeat="appointment in appointments | orderBy:['noshow', 'timeIn == null', 'timeReturnedPapers == null', 'timeAppointmentStarted == null', 'scheduledTime'] | searchFor: clientSearch"
+					ng-repeat="appointment in appointments | orderBy:['noShow', 'timeIn == null', 'timeReturnedPapers == null', 'timeAppointmentStarted == null', 'scheduledTime'] | searchFor: clientSearch"
 					ng-if="appointment.completed == null"
 					ng-click="selectClient(appointment)">
 					<th class="queue-name" data-header="Name">{{appointment.firstName}} {{appointment.lastName}}</th>
 					<td class="queue-status" data-header="Progress">
-						<span class="pill pill-noshow" ng-if="appointment.noshow">No-show</span>
-						<span ng-if="!appointment.noshow">
+						<span class="pill pill-noshow" ng-if="appointment.noShow">No-show</span>
+						<span ng-if="!appointment.noShow">
 							<span class="pill" ng-class="appointment.checkedIn ? 'pill-complete': 'pill-incomplete'">Checked In</span>
 							<span class="pill" ng-class="appointment.paperworkComplete ? 'pill-complete': 'pill-incomplete'">Completed Paperwork</span>
 							<span class="pill" ng-class="appointment.preparing ? 'pill-complete': 'pill-incomplete'">Preparing</span>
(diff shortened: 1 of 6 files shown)


server/queue/queue.php
```diff
@@ -15,11 +15,12 @@ function loadPublicQueue($data) {
 	GLOBAL $DB_CONN;
 	$query = "SELECT Appointment.appointmentId, TIME_FORMAT(scheduledTime, '%l:%i %p') AS scheduledTime, firstName, lastName, 
 				timeIn, timeReturnedPapers, timeAppointmentStarted, timeAppointmentEnded, completed,
-				(DATE_ADD(AppointmentTime.scheduledTime, INTERVAL 15 MINUTE) < NOW() AND timeIn IS NULL) AS noshow 
+				(DATE_ADD(AppointmentTime.scheduledTime, INTERVAL 15 MINUTE) < NOW() AND timeIn IS NULL) AS noShow,
+				(AppointmentTime.scheduledTime < Appointment.createdAt) AS walkIn
```
MatthewMeacham: I would love a comment on this design. I'm not certain running this comparison in the SQL statement is the best idea? We could easily do this on the front-end or in PHP instead if we returned the unformatted scheduledTime and createdAt. Thoughts from a best practices perspective?

queue/private.php
```diff
@@ -39,13 +39,14 @@
...
 				<tr class="pointer"
-					ng-repeat="appointment in appointments | orderBy:['noshow', 'timeIn == null', 'timeReturnedPapers == null', 'timeAppointmentStarted == null', 'scheduledTime'] | searchFor: clientSearch"
+					ng-repeat="appointment in appointments | orderBy:['noShow', 'timeIn == null', 'timeReturnedPapers == null', 'timeAppointmentStarted == null', 'scheduledTime', 'walkIn'] | searchFor: clientSearch"
 					ng-if="appointment.completed == null"
 					ng-click="selectClient(appointment)">
 					<th class="queue-name" data-header="Name">{{appointment.firstName}} {{appointment.lastName}}</th>
 					<td class="queue-status" data-header="Progress">
-						<span class="pill pill-noshow" ng-if="appointment.noshow">No-show</span>
-						<span ng-if="!appointment.noshow">
+						<span class="pill pill-no-show" ng-if="appointment.noShow">No-show</span>
```
jjstobbe: Nice renaming from `noshow` to `noShow`, incredible work.

queue/queueController.js
```diff
@@ -29,7 +29,8 @@ define('queueController', [], function() {
 						appointment.preparing = appointment.timeAppointmentStarted != null;
 						appointment.ended = appointment.timeAppointmentEnded != null;
 						appointment.name = appointment.firstName + " " + appointment.lastName;
-						appointment.noshow = appointment.noshow == true; // Do this since the SQL returns 0/1 for false/true, and we want it to be true/false instead of 0/1
+						appointment.noShow = appointment.noShow == true; // Do this since the SQL returns 0/1 for false/true, and we want it to be true/false instead of 0/1
```
jjstobbe: I've confused about what this is doing. This seems like it shouldn't do anything, but apparently it doesn't work without it so 🤷‍♂️

PR #257, opened 2018-11-23
[77] Shift Management

MatthewMeacham:
This PR creates a new "Site Management" page where a site administrator can create shifts at a site, and create appointment times at a site. In the future, this page will be expanded to include the functionality for editing, adding, and deleting sites. 

## Testing
1. Run `./scripts/build.sh`
2. Run the DB scripts
3. Log in as preparer
    * You should not be able to see the "Site Management" link under the "Management" tab
4. Logout and login as siteadmin
    * You should be able to see the "Site Management" link under the "Management" tab
5. Navigate to the "Site Management" page
    * You should see all 7 sites listed on the left-hand side of the screen
6. Click on "Test Site"
    * Information for "Test Site" should load and be displayed
    * You should see 2 volunteer shifts and 7 appointment times listed
7. Hit the "Add Shift" button
    * A form should show
        * Hitting the cancel button should hide the form
    * Choose a date
        * You should not be able to choose a date in the past
    * Choose a start time
    * Choose an end time
        * The end time should start 30 minutes after the start time
        * If you change the start time, the end time should clear and you will be re-prompted to select a time
    * Hit Save
        * You should get a green success notification
        * The shift should be added to the table of shifts above
8. Refresh the page
    * The shift you just added should be in that table for "Test Site"
9. Hit the "Add Appointment Time" button
    * A form should show
        * Hitting cancel should hide the form
    * You should only be able to select a date that has a shift on it (I suggest choosing the date of the shift you just added above!)
    * You should only be able to choose a time that is within the shifts on that day
        * Note that if there is already an appointment time at a given time, you can't schedule another appointment time then
    * Changing the approximate length in minutes field should update the scheduled times you can choose
    * Hit the Save button
        * You should get a green success notification
        * The appointment time should be added to the table of appointment times above
10. Refresh the page
    * The appointment time you just added should be in that table for "Test Site"
11. Navigate to the profile page
    * Sign up as a preparer for the shift that you just added
        * You should be able to successfully sign up
12. Log out 
13. Navigate to the appointment sign up page
    * NOTE: You will need to remove the comments that are preventing you from being able to sign up for an appointment
    * Sign up for an appointment at the appointment time that you just created (you should be able to do this since you signed up a preparer for this shift time)
        * You should be able to sign up successfully

If all that works, then this piece of functionality should be good!


assets/js/utilities/notificationUtilities.js
```diff
@@ -4,9 +4,9 @@ define('notificationUtilities', [], function() {
 		return {
 			giveNotice: function(title, message, affirmative = true) {
 				WDN.initializePlugin('notice');
-				const body = angular.element(document.querySelector('body' ));
+				const body = angular.element(document.querySelector('body'));
 				body.append(`
-					<div class="wdn_notice ${affirmative ? 'affirm' : 'negate'}" data-overlay="maincontent" data-duration="10">
```
MatthewMeacham: @CollinsSpencer I removed the `data-duration` since it was suggested not to use by the WDN page. Also, I added fixed positioning so that one wouldn't need to jump the page up to the top so a user could see the notification. Thoughts on if this was a good idea?

management/sites/sites.php
```diff
@@ -0,0 +1,318 @@
...
+							<div class="form-textfield">
+								<label class="form-label form-required">Date</label>
+								<input type="text" 
+									id="addShiftDateInput" 
+									name="dateInput" 
+									placeholder=" -- Select a Date -- " 
+									autocomplete="off"
+									required>
+								<div ng-show="form.$submitted || form.dateInput.$touched">
+									<label class="error" ng-show="form.dateInput.$error.required">This field is required.</label>
```
CollinsSpencer: FYI, although you can't select a prior date, you can still change the text to a prior date (and it does create a new shift on that past date without any errors). Not necessarily a bug unless we really don't want them to create new shifts on past dates. Shouldn't create any problems though if they do for some odd reason.

management/sites/sitesController.js
```diff
@@ -0,0 +1,311 @@
...
+			};
+
+			let times = [];
+
+			// Get the possible times within each shift
+			for (const shift of shifts) {
+				const startTime = getTime(shift.startTime);
+				const endTime = getTime(shift.endTime);
+
+				const generatedTimes = generateTimes(timeInterval, startTime, endTime);
```
CollinsSpencer: I think for the third parameter in `generateTimes` really depends on if the times need to fall completely within the start and end times or if they just need to start in that range. For the appointment times, we want those to fall entirely within the shift. I think we should use `endTime - timeInterval` instead of just `endTime`. It correctly throws an error, but I feel like we should just not show that option.

management/sites/sites.php
```diff
@@ -0,0 +1,318 @@
...
+										<a class="tooltip pointer" title="This is the approximate length in minutes of an appointment. This number is critical to determining how many preparers are available at a certain time. Typically this number will be 60.">
+											<span class="wdn-icon-info" aria-hidden="true"></span>
+										</a>
+									</th>
+								</tr>
+							</thead>
+							<tbody>
+								<tr ng-repeat="appointmentTime in siteInformation.appointmentTimes">
+									<th id="appointmentTime{{appointmentTime.appointmentTimeId}}">{{appointmentTime.scheduledTimeString}}</div>
+									<td headers="appointmentTimeMinimumNumberOfAppointmentsHeader appointmentTime{{appointmentTime.appointmentTimeId}}">{{appointmentTime.minimumNumberOfAppointments == null || appointmentTime.minimumNumberOfAppointments == 0 ? 'N/A' : appointmentTime.minimumNumberOfAppointments}}</td>
```
CollinsSpencer: Any reason for showing N/A rather than just 0 here?

management/sites/sitesController.js
MatthewMeacham: It wasn't enough to use `endTime - timeInterval` on this because if a shift time went from 2pm-4pm and we have appointments that are 60 minutes long, then we wanted to have both a 2pm and 3pm appointment times. 

So what I did was use `endTime - timeInterval` here and then have a `endTimeInclusive` parameter that defaults to false in that `generateTimes` method, and when it is set to true, the method will generate a time for the `endTime` part as well.. Thoughts on if this is a good approach?

Another option that would've worked was using `endTime - timeInterval + 1` instead of any of the other changes :p

management/sites/sites.php
MatthewMeacham: I'm partial to letting this one slide as is.

assets/js/utilities/notificationUtilities.js
CollinsSpencer: I think this is fine.

management/sites/sites.php
CollinsSpencer: Sounds good.

management/sites/sitesController.js
CollinsSpencer: Ah. Makes sense. The changes look good.

management/sites/sites.php
CollinsSpencer: I think "N/A" is fine. That way it matches the Max Appts column too.

management/sites/sites.php
```diff
@@ -0,0 +1,318 @@
...
+									</th>
+									<th id="appointmentTimeMaximumNumberOfAppointmentsHeader">
+										Max Appts
+										<a class="tooltip pointer" title="This is the maximum number of appointments for this time slot, meaning even if there are more preparers than this number, this is the number of appointments that will be allowed to be scheduled. Default is N/A.">
+											<span class="wdn-icon-info" aria-hidden="true"></span>
+										</a>
+									</th>
+									<th id="appointmentTimePercentageAppointmentsHeader">
+										Percentage Appts
+										<a class="tooltip pointer" title="This specifies the percentage of the preparers that are allotted for appointments, that is: number of appointments allowed to be scheduled = number of preparers * percentage appointments. Default is 100%.">
```
CollinsSpencer: While it is all correct information, I'm not sure that this description will be clear to the users. I believe the purpose of the percentage appts value is to only allow a certain percentage of the appointments to be scheduled online, thus reserving the other appointments for walk-ins. I wonder if we should hint that the usage/purpose of this value is to save some the time of some preparers for walk-ins. I'm thinking something along these lines...

"This specifies the percentage of the preparers that are allotted for scheduled (as opposed to walk-in) appointments. That is: number of appointments allowed to be scheduled online = number of preparers * percentage appointments. Default is 100%."

Thoughts?

MatthewMeacham: Good idea, I'll make that change. On this note though, in my talk with Dr. Moody and Katie, this year, they want appointment times to be on the hour (i.e 1pm, 2pm, 3pm), and then have 2 appointments allowed to be scheduled per preparer... I was planning on using the `percentageAppointments` functionality to achieve this result. Do you have any thoughts on this?

CollinsSpencer: Interesting. So they want to intentionally overbook? In terms of using `percentageAppointments` to achieve this, are you thinking we would allow the value to be set above 100%? Do we just want to increase the upper limit to something like 300%?

MatthewMeacham: Yah they do want to intentionally overbook, but they made it sound like appointments weren't taking a full hour in most cases, but they didn't want to shorten appointment times to 30 minutes because on the hour gave them more flexibility. So my thought was to set `percentageAppointments` to 200% on a lot of the appointment times in order to achieve this. 

So my initial thought was just to remove the upper limit constraint on the `percentageAppointment` field, but I think an upper limit like 300% is more appropriate. You think this is a "good" way to achieve the result we're after? At least initially to see if it is something that even works out for them?

PR #259, opened 2018-11-24
Update the years on the questions

MatthewMeacham:
Only creating a PR because I want your thoughts--these years need to be incremented by 1 every year. Should we just use JavaScript to get the client's current year? I realize it is unreliable, but I think that would still be better than having to come in here every year and increment these dates...

### DIFF
--- a/signup/signup.php
+++ b/signup/signup.php
@@ -1,267 +1,276 @@
+<?php
+	$today = date('Y-m-d');
+	$taxDay = date('Y-04-15');
+	$taxYear = ($today > $taxDay) ? date('Y', strtotime('+1 year')) : date('Y');
+?>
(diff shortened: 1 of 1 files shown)


CollinsSpencer:
Since we are currently setting this text in the php, it would probably be best to just set the date on the server. My initial thoughts would be to get the year when a certain date next occurs and then base all of these off of that year.

e.g.
```php
$today = date('Y-m-d');
$taxDay = date('Y-04-15');
if ($today > $taxDay) {
  $taxDay = date('Y-04-15', strtotime('+1 year'));
}
```

MatthewMeacham:
Good idea @CollinsSpencer, made those changes

PR #263, opened 2018-11-26
[261] Enable Admin to Edit Volunteer Information

MatthewMeacham:
This PR enables an admin to edit all information of a user by hitting edit on the user management page. 

Please note that the `/management/users/` page uses an extremely old pattern of creating stuff in the PHP response and JQuery. I wanted to update this to use AngularJS and whatnot, but it would've been way too much work for the scope of this PR. I did slightly refactor the JS file though to be a little easier to work with, still not great, but it is better.

Also, I want specific comment on whether or not an admin should be allowed to edit the email of a user. Currently, users can't even edit their own emails. This specific feature request stemmed from a time when I had to change the email address of a user in the DB because an admin had accidentally typed it wrong when they added the user. However, I'm not sure enabling editing of the email without some sort of email verification (like happens during user registration) is a good idea. But I do want opinions on it. @CollinsSpencer @jjstobbe 

## Testing
1. Run the DB scripts
2. Run the `./scripts/build.sh`
3. Log in as `siteadmin@test.test`
4. Navigate to the "User Management" page
5. Hit the edit button on any user
    * A modal should pop up
        * The modal should be populated with the information of the selected user
        * Hitting cancel, the X, or outside of the modal should close the modal
6.  Change any/all of the information and hit the "Save" button
    * It should save successfully, close the modal, and the table should update
7. Change the information on another user
    * It should save successfully and not affect any other users

## Additional Testing
1. Ensure the "Add User" functionality still works properly


management/users/users.js
```diff
@@ -12,77 +12,88 @@ WDN.initializePlugin('modal', [function() {
...
-	
+					let isValid = true;
+
 					$(this).find('input').each(function(){
-						var noValue = !$(this).val();
-						if(noValue){
-							valid = false;
+						const noValue = !$(this).val();
+						if (noValue) {
+							isValid = false;
```
jjstobbe: Could do something like this and get rid of the if statement:
```
isValid = !!$(this).val();
```
I'm not convinced this is better though, so it's up to you..

management/users/users.js
```diff
@@ -91,32 +102,32 @@ WDN.initializePlugin('modal', [function() {
...
 							addPermissionArr: addPermissionArr
 						},
 						success: function(response){
-							if(response.success){
-								refreshUserTable();
-							}else{
-								refreshUserTable();
-								alert(response.error);
+							if (!response  || !response.success) {
+								alert(response.error || 'There was an error updating the permissions. Please refresh the page and try again');
```
jjstobbe: Not alerts, oh man 😢

management/users/users.js
```diff
@@ -125,66 +136,181 @@ WDN.initializePlugin('modal', [function() {
...
+							$(this).off();
+						} else {
+							$('#edit-user-form button[type=submit]').prop('disabled', false);
+						}
+					});
+				});
+			};
+
+			function getUserInformation(userId, callbackFunction) {
+				$.ajax({
```
jjstobbe: Instead of passing around callbacks, a cleaner solution would be javascript promises.
You can essentially just return the ajax call, then do `.then` to get the response.

```
return $.ajax({
		dataType: 'json',
		method: 'GET',
		url: '/server/management/users/users.php',
		data: {
			action: 'getUserInformation',
		userId: userId
	})
       .then((response) => {
           if (!response || !response.success) {
                 alert('whatever')
                 return
           }
           return response.user;
       })
```
then back when you call it..
```
getUserInformation(userId)
      .then((userData) => {
           $('#editFirstName').val(userData.firstName);
           // ... the rest of them
      })
```

management/users/users.js
```diff
@@ -125,66 +136,181 @@ WDN.initializePlugin('modal', [function() {
...
+			function refreshUserTable() {
+				$.ajax({
+					dataType: 'json',
+					method: 'POST',
+					url: '/server/management/users/users.php',
+					data: {
+						action: 'getUserTable'
+					},
+					success: function(response) {
+						if (!response || !response.success) {
```
jjstobbe: You do a ton of these checks, might be worth making a helper function to put this logic in one place.
```
checkSuccessful(response, errorMessage = 'There was an error fetching the data. Please refresh the page.') {
    isSuccessful = response && response.success; // Or whatever it is
    if (!isSuccessful) {
        notify(errorMessage);
    }
    return isSuccessful;
}
```

PR #265, opened 2018-12-27
2019 Tax Season Data Script

MatthewMeacham:
This is the start of the script that will add all the data for the 2019 tax season to the production database. I opted to create a PHP script this year since it was a lot less work than a long SQL script. 

## PROD DB Evolutions
The following query will need to be run against the prod database before the script could be run.   
Query 1: Archive the Jackie Gaughan Site. 
```sql
UPDATE Site 
SET archived = TRUE
WHERE siteId = 3; # Note I pulled this id straight from the PROD DB
```
Query 2: Add the percentageAppointment check to the DB
```sql
ALTER TABLE AppointmentTime
ADD CONSTRAINT percentageCheck CHECK (percentageAppointments>=0 AND percentageAppointments<=300);
```


mysql_scripts/table_creation/FullDB.sql
```diff
@@ -80,7 +80,7 @@ CREATE TABLE AppointmentTime (
 	maximumNumberOfAppointments INTEGER UNSIGNED DEFAULT NULL,
 	percentageAppointments INTEGER UNSIGNED NOT NULL DEFAULT 100,
 	approximateLengthInMinutes INTEGER UNSIGNED NOT NULL DEFAULT 60,
-	CONSTRAINT percentageCheck CHECK (percentageAppointments>=0 AND percentageAppointments<=100),
+	CONSTRAINT percentageCheck CHECK (percentageAppointments>=0 AND percentageAppointments<=300),
```
MatthewMeacham: Hey funny story, so this constraint actually just doesn't exist in the PROD DB. Should we remove it from our table creation script? Or should I add it to the PROD DB? @CollinsSpencer

server/scripts/2019TaxSeasonDataScript.php
```diff
@@ -0,0 +1,361 @@
...
+
+}
+
+function insertInternationalStudentScholarSiteData() {
+	$dataAlreadyInserted = true;
+	if ($dataAlreadyInserted) {
+		die('The ISS data has already been inserted');
+	}
+
+	// TODO: NEED TO FINISH UPDATED STATION LIMITING CODE BEFORE I CAN ENTER THIS
```
MatthewMeacham: Actually, I will probably go ahead and put in the data for all the other sites through this script, and then come back to this later--international appointments don't start until March sometime, so we have time to not put in the data now.

mysql_scripts/table_creation/FullDB.sql
MatthewMeacham: Good point. I will add it to the DB, and update the front-end and back-end to match the 300% max.

PR #267, opened 2018-12-28
Various fixes for going into future years

MatthewMeacham:
There were a couple spots that I discovered on a brief peruse of the website that were going to grab incorrect information as we go into the second year of this application. This PR addresses those small concerns.


server/profile/profile.php
```diff
@@ -69,18 +69,21 @@ function getShifts() {
 	GLOBAL $USER, $DB_CONN;
 	$userId = $USER->getUserId();
 
+	date_default_timezone_set('America/Chicago');
```
CollinsSpencer: Good call.

PR #301, opened 2019-12-27
Update NPM Dependencies

MatthewMeacham:
Updated the NPM dependencies so that @dependabot would stop opening PRs.


MatthewMeacham:
Holding off on merging. It appears that this broke production building (the CSS files are no longer processed/hashed). Will investigate which NPM package broke that functionality and get back to this.

MatthewMeacham:
It looks like I fixed the issue. Postcss had changed one of the command line arguments (it no longer accepted `-b` as short for `--base`). I've corrected that and now it appears to be working properly.

PR #302, opened 2019-12-28
Update Remaining Appointment Spots Calculation

MatthewMeacham:
Updated how we calculate remaining appointment spots. I don't think the logic was complete before, but I believe this completes it with both correct maximum and correct minimum logic.

### DIFF
--- a/server/api/appointments/getTimes.php
+++ b/server/api/appointments/getTimes.php
@@ -112,11 +112,15 @@ function getInternationalAppointmentTimes($year, $after, $treatyType) {
 }
 
 function calculateRemainingAppointmentsAvailableForResidentialAppointment($appointmentCount, $percentAppointments, $preparerCount, $minimum, $maximum) {
-	if (isset($maximum)) {
-		$availableAppointmentSpots = $maximum;
-	} else {
-		$availableAppointmentSpots = max($minimum, $preparerCount);
+	$availableAppointmentSpots = $preparerCount;
+	if (isset($minimum)) {
+		$availableAppointmentSpots = max($minimum, $availableAppointmentSpots);
 	}
+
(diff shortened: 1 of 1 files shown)


server/api/appointments/getTimes.php
```diff
@@ -112,11 +112,15 @@ function getInternationalAppointmentTimes($year, $after, $treatyType) {
...
+	}
+
 	if (isset($maximum)) {
-		$availableAppointmentSpots = $maximum;
-	} else {
-		$availableAppointmentSpots = max($minimum, $preparerCount);
+		$availableAppointmentSpots = min($maximum, $availableAppointmentSpots);
 	}
+
 	$availableAppointmentSpots *= $percentAppointments / 100;
```
jcarrigan2: If $percentAppointments should always be 100 for us, why not comment this out? I could see me accidentally putting an extra 0 on the 100 for percent. This is more just for practice for me suggesting a change tbh.

MatthewMeacham: Good question -- it's not always 100%, it might be this year, but this value has a lot of power behind it.

Originally, this was added because they wanted to do 75% appointments and allow 25% for walk-ins. This percentage controlled that. 

Later on, they wanted to do 2 appointments per preparer per hour. We were then able to use this percentAppointments to accomplish that (by setting it to 200%). 

So, there's a lot of power behind this variable. To your point though about accidentally adding an extra 0... There's actually a DB constraint on this variable that only allows to go up to 300% (for the exact reason you mentioned), so that ideally shouldn't be a problem. 

While the functionality may not be used right now, I would be partial to leaving this in since it's perhaps the easiest way for us/VITA admins to control appointments.

PR #306, opened 2020-01-10
Update 2020TaxSeasonDataScript.php

jcarrigan2:
Added the Nebraska Union and International Student Scholar Sites.
Assumed the limit on the number of greeters and site coordinators was 1

### DIFF
--- a/server/scripts/2020TaxSeasonDataScript.php
+++ b/server/scripts/2020TaxSeasonDataScript.php
@@ -24,6 +24,9 @@ function insert2020Data() {
 	// insertBennettMartinLibraryData();
 	// insertFStreetCommunityCenterData();
 	// insertSoutheastCommunityCollegeData();
+	// insertInternationalStudentScholarSiteData();
+	// insertNebraskaUnionData();
+
 	die('SUCCESS');
 }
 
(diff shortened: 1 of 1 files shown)


server/scripts/2020TaxSeasonDataScript.php
```diff
@@ -302,6 +305,122 @@ function insertSoutheastCommunityCollegeData() {
...
+
+	try {
+		$DB_CONN->beginTransaction();
+
+		// Tuesdays
+		$tuesdayDates = array('2020-03-03', '2020-03-10', '2020-03-17', '2020-03-31');
+		foreach ($tuesdayDates as $date) {
+			$shiftId = insertShift("$date 13:00:00", "$date 18:00:00", $siteId);
+
+			$firstAppointmentTimeId = insertAppointmentTime("$date 13:00:00", 41, 41, 100, 60, $siteId);
```
MatthewMeacham: Not sure it's an issue--but technically we don't need a min/max appointments on these since it's hard-coded in the back-end rn unfortunately. So those values aren't used, but I think they can stay

server/scripts/2020TaxSeasonDataScript.php
```diff
@@ -302,6 +305,122 @@ function insertSoutheastCommunityCollegeData() {
...
+	GLOBAL $DB_CONN;
+
+	$siteCoordinatorRoleId = 1;
+	$greeterRoleId = 2;
+	$preparerRoleId = 3;
+	$reviewerRoleId = 4;
+	
+	$dataAlreadyInserted = true;
+	if ($dataAlreadyInserted) {
+		die('The AL data has already been inserted');
```
MatthewMeacham: This should be "The Nebraska Union data..."

server/scripts/2020TaxSeasonDataScript.php
```diff
@@ -302,6 +305,122 @@ function insertSoutheastCommunityCollegeData() {
...
+		}
+
+		// Default Site Role Limits
+		insertSiteRoleLimit(1, $siteCoordinatorRoleId, $siteId);
+		insertSiteRoleLimit(1, $greeterRoleId, $siteId);
+
+		$DB_CONN->commit();
+	} catch (Exception $e) {
+		$DB_CONN->rollback();
+		throw new Exception('Failed inserting Anderson Library data', MY_EXCEPTION);
```
MatthewMeacham: Similarly here, should be Nebraska Union

PR #308, opened 2020-01-11
Appointment Analytics

MatthewMeacham:
This is something I've just wanted to do for a while, especially now that we actually have more than one season worth of data.  

This PR creates an "Analytics" page that an admin can look at ("can use admin tools" permission required. It looks like this right now:
image

It uses ChartJS for the charting, and it shows:
* Number of appointments each year (broken down by residential/international)
* Number of UNL/Wesleyan appointments each year
* Number of appointments each year per site

### DIFF
--- a/management/analytics/analyticsController.js
+++ b/management/analytics/analyticsController.js
@@ -1,0 +1,74 @@
+define('analyticsController', [], function() {
+
+	function analyticsController($scope, AnalyticsDataService, NotificationUtilities) {
+
+		$scope.sites = [];
+
+		$scope.loadSites = () => {
+			AnalyticsDataService.getSites().then((result) => {
+				if (result == null) {
+					NotificationUtilities.giveNotice('Failure', 'Unable to load the sites.', false);
+					return;
+				}
+
+				$scope.sites = result;
+			});
+		};
+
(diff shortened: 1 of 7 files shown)


management/analytics/analyticsController.js
```diff
@@ -0,0 +1,74 @@
...
+				$scope.sites = result;
+			});
+		};
+
+		// Number of appointments scheduled
+		// Number of UNL/Wesleyan appointments scheduled
+		// Number of international appointments scheduled
+		
+
+		$scope.addDemoData = () => {
```
jcarrigan2: what are these comments/whitespace for?

MatthewMeacham: Ah good catch--they're for nothing. Just trying to keep stuff straight when I was writing this. I'll delete these.

MatthewMeacham:
@CollinsSpencer haha I'm not a fan of the random colors either--do you have a good idea of what to do? I could always just specify like 10 colors in an array, and just index into that color array... But what colors should I pick?

MatthewMeacham:
@CollinsSpencer Used a color palette. It now looks like this:
image

PR #310, opened 2020-01-12
Automatically Send Confirmation Email

MatthewMeacham:
After signing up, a confirmation email is automatically sent to the client. This was a request that Katie mentioned that clients wanted--it also was a problem where a volunteer would sign someone up, but they wouldn't receive a confirmation email since the volunteer wouldn't hit the "Email me" button at the bottom. This addresses that issue.

I also made a small styling change to the `form.css` to have `required` after a required input rather than `*`. I saw this on the WDN website in their form, and wanted to move to be more in-line with that. I also think it looks a little nicer:
image

### DIFF
--- a/assets/css/form.css
+++ b/assets/css/form.css
@@ -3,9 +3,11 @@
 }
 
 label.form-required::after {
-	content: " *";
+	content: " required";
 	color: #d00000;
-	font-size: 1.247rem;
+	font-style: italic;
+	font-size: 0.75rem;
+	vertical-align: super;
 }
 
 [data-toggle=buttons]>.dcf-btn-group>.dcf-btn input[type=checkbox],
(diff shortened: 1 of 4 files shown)


assets/css/form.css
```diff
@@ -3,9 +3,11 @@
 }
 
 label.form-required::after {
-	content: " *";
+	content: " required";
```
CollinsSpencer: I like this change. 👍

signup/signupController.js
```diff
@@ -127,15 +128,12 @@ define('signupController', [], function() {
...
-						NotificationUtilities.giveNotice('Success', 'The email has been sent!');
+						NotificationUtilities.giveNotice('Success', 'A confirmation email has been sent!');
 					} else {
-						$scope.emailButton.disabled = false;
 						NotificationUtilities.giveNotice('Failure', response.error, false);
 					}
 				} else {
-					$scope.emailButton.disabled = false;
-					NotificationUtilities.giveNotice('Failure', 'There was an error on the server! Try again or please print this page instead.', false);
+					NotificationUtilities.giveNotice('Failure', 'There was an error sending a confirmation email! Please print this page instead.', false);
```
jcarrigan2: Good change!

PR #314, opened 2020-01-14
Quick Fix for Less Appointments for International Site

MatthewMeacham:
Katie wanted the last time slot at the international sites to have less appointments than the other times... Unfortunately, the international system hasn't really been built out enough to handle this situation. So this is a quick fix that directly targets those 5pm appointment times by ID and halves the number of available appointments.

### DIFF
--- a/server/api/appointments/getTimes.php
+++ b/server/api/appointments/getTimes.php
@@ -125,16 +125,30 @@ function calculateRemainingAppointmentsAvailableForResidentialAppointment($appoi
 	return ceil($availableAppointmentSpots) - $appointmentCount;
 }
 
-function calculateRemainingAppointmentsAvailableForInternationalAppointment($treatyType, $numberAppointmentsAlreadyScheduled) {
-	if ($treatyType === 'china') {
-		return 15 - $numberAppointmentsAlreadyScheduled;
-	} else if ($treatyType === 'india') {
-		return 6 - $numberAppointmentsAlreadyScheduled;
-	} else if ($treatyType === 'treaty') {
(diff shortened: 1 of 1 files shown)


server/api/appointments/getTimes.php
```diff
@@ -125,16 +125,30 @@ function calculateRemainingAppointmentsAvailableForResidentialAppointment($appoi
...
-		return 5 - $numberAppointmentsAlreadyScheduled;
-	} else if ($treatyType === 'non-treaty') {
-		return 15 - $numberAppointmentsAlreadyScheduled;
-	} 
+function calculateRemainingAppointmentsAvailableForInternationalAppointment($treatyType, $numberAppointmentsAlreadyScheduled, $appointmentTimeId) {
+	// TODO: This is a real band-aid that we are using to enable less appointments for certain time slots
+	$appointmentTimeIdsForLessAppointments = array(968, 969, 970, 971, 972);
+	if (in_array($appointmentTimeId, $appointmentTimeIdsForLessAppointments)) {
+		if ($treatyType === 'china') {
+			return 7 - $numberAppointmentsAlreadyScheduled;
```
jcarrigan2: Do you care if these return negative numbers?

jcarrigan2: Do you care if these return negative numbers?

MatthewMeacham: Yah that's fine--the FE won't let clients schedule into times that are negative, and it actually informs a logged-in volunteer that the time is "overscheduled by X appointments"

PR #332, opened 2020-05-01
Additional virtual appointment changes

MatthewMeacham:
A few more changes for virtual appointment preparation. These changes are:
* Added a new field to appointment sign-up "Best time to Call"
* Added a checkbox to appointment sign-up saying "I agree to have my tax return prepared virtually"
* Added a checkbox to upload documents that also says "I agree to have my tax return prepared virtually"
* Added some additional text saying clients must save the fillable PDFs on upload documents in order for their information to save

These changes do require a DB evolution. The follow script will need to be run on the PROD DB before these changes are deployed:
```sql
ALTER TABLE Client
    ADD COLUMN bestTimeToCall VARCHAR(255) NULL;
```

### DIFF
--- a/signup/signup.php
+++ b/signup/signup.php
@@ -3,22 +3,13 @@
 	$today = date('Y-m-d');
 	$dateAppointmentSignUpsStart = date('Y-04-27');
 	$taxDay = date('Y-07-15');
-	$overrideForCovid19 = false;
 	
 	$taxYear = ($today > $taxDay) ? date('Y', strtotime('+1 year')) : date('Y');
 ?>
 
 <!-- Appointment Signup with no success -->
 <div class="dcf-pb-7" ng-if="successMessage == null">
-	<?php if ($overrideForCovid19) { ?>
-		<h4>
-			In response to COVID-19, all VITA tax appointments were canceled starting March 16th. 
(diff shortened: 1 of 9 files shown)


signup/signup.php
```diff
@@ -298,9 +297,15 @@ class="dcf-input-select"
 
 			<div appointment-picker class="dcf-mb-5"></div>
 
+			<div class="dcf-input-checkbox">
+				<input id="agree-to-virtual-preparation-checkbox" type="checkbox" ng-model="agreeToVirtualPreparationCheckbox.checked" value="false">
+				<label for="agree-to-virtual-preparation-checkbox">I agree to have my tax return prepared virtually. See <a href="https://www.irs.gov/pub/irs-pdf/f14446.pdf" target="_blank">Form 14446 (Virtual VITA/TCE Taxpayer Consent)</a></label>
```
CollinsSpencer, line 302: Any reason for using this version of the form rather than the one that we have on the server with the fields already set up?

MatthewMeacham: Tbh I just didn't want to copy the method for downloading the file over to this. What are your thoughts? The link Katie sent me was just to the IRS website, she never mentioned re-using the one that we have the clients filled out

CollinsSpencer: Makes sense. I'm fine with it how it is, but I was just curious if there was a reason for it being different.

PR #336, opened 2020-05-13
No particular time for virtual appointments

MatthewMeacham:
This code adds a simple `isVirtual` flag to the `Site` table and uses that to change the time options during appointment sign up. This PR also removes the hard-coded pieces for virtual appointment in the appointment confirmation utility.

# DB Evolution
The following SQL script will need to be run in PROD before this change is deployed
```sql
ALTER TABLE Site
    ADD COLUMN isVirtual BOOLEAN NOT NULL DEFAULT FALSE;
UPDATE Site
    SET isVirtual = TRUE
    WHERE siteId = --virtual site ID
```


components/appointmentPicker/appointmentPickerSharedPropertiesService.js
```diff
@@ -8,6 +8,7 @@ define('appointmentPickerSharedPropertiesService', [], function() {
 			selectedDate: null,
 			selectedSite: null,
 			selectedSiteTitle: null,
+			isSelectedSiteVirtual: false,
```
jcarrigan2, line 11: Why did you decide to initialize this as false?

server/utilities/appointmentConfirmationUtilities.class.php
```diff
@@ -5,8 +5,6 @@
 class AppointmentConfirmationUtilities {
 
 	public static function generateAppointmentConfirmation($appointmentId) {
-		// TODO: THIS SHOULD NOT BE HARD-CODED TO TRUE AFTER VIRTUAL APPOINTMENTS STOP
```
jcarrigan2, line 8: Love to see the TODOs getting deleted. Great change to the isVirtual field!

components/appointmentPicker/appointmentPickerSharedPropertiesService.js
MatthewMeacham: I guess it just seemed like a sane default for this--in most scenarios, I imagine the site will be physical rather than virtual. I just as easily could've done `null`, but I think since this one is a bit more deterministic, `false` suited better. But this is just me rationalizing it now--in the moment, I didn't think of any of this.

server/utilities/appointmentConfirmationUtilities.class.php
MatthewMeacham: Yah it feels good to cross one of these off haha

PR #337, opened 2020-05-14
Add Unique URLs to Appointment Management Page

MatthewMeacham:
Just added the self-service appointment reschedule URL and the upload documents URL to the appointment management page so that a volunteer could fetch these for a client if the client didn't save the information.

### DIFF
--- a/server/management/appointments/appointments.php
+++ b/server/management/appointments/appointments.php
@@ -49,7 +49,7 @@ function getAppointments($year) {
 				JOIN AppointmentTime ON Appointment.appointmentTimeId = AppointmentTime.appointmentTimeId
 				JOIN Site ON AppointmentTime.siteId = Site.siteId
 				JOIN Client ON Appointment.clientId = Client.clientId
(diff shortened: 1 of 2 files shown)


server/management/appointments/appointments.php
```diff
@@ -49,7 +49,7 @@ function getAppointments($year) {
 				JOIN AppointmentTime ON Appointment.appointmentTimeId = AppointmentTime.appointmentTimeId
 				JOIN Site ON AppointmentTime.siteId = Site.siteId
 				JOIN Client ON Appointment.clientId = Client.clientId
-				JOIN SelfServiceAppointmentRescheduleToken ON Appointment.appointmentId = SelfServiceAppointmentRescheduleToken.appointmentId
+				LEFT JOIN SelfServiceAppointmentRescheduleToken ON Appointment.appointmentId = SelfServiceAppointmentRescheduleToken.appointmentId
```
jcarrigan2, line 52: What's the reason for this change? Thanks~

MatthewMeacham: I needed the `token` in order to construct the unique URLs. The left join is because not every appointment has a token (in particular, the appointments that were created before the token infrastructure was built or ones that were created manually by us rather than through the site), but we still want all the appointments.

PR #340, opened 2020-05-28
Email Input Required for Virtual Appointments

MatthewMeacham:
Made the email input required for virtual appointments. This is a request from Katie--it's easier to reach out to clients via email rather than calling them for these virtual appointments.

Also added some text saying the form is invalid when the form is submitted and contains errors.

### DIFF
--- a/signup/signupController.js
+++ b/signup/signupController.js
@@ -168,7 +168,7 @@ define('signupController', [], function() {
 		};
 
 		$scope.isEmailRequired = () => {
-			return $scope.sharedProperties.isSelectedSiteVirtual == true;
+			return $scope.sharedProperties.isSelectedSiteVirtual === true;
 		};
 
 		$scope.downloadForm14446 = () => {
--- a/signup/signupController.js
+++ b/signup/signupController.js
@@ -167,6 +167,10 @@ define('signupController', [], function() {
 			}
 		};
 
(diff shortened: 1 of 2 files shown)


signup/signupController.js
```diff
@@ -167,6 +167,10 @@ define('signupController', [], function() {
 			}
 		};
 
+		$scope.isEmailRequired = () => {
+			return $scope.sharedProperties.isSelectedSiteVirtual == true;
```
CollinsSpencer, line 171: Could we switch this from `==` to `===`?

MatthewMeacham: We could.. I opted for the `==` because I know behind-the-scenes the `isSelectedSiteVirtual` property comes from the `isVirtual` field in the DB, and I didn't want to risk a situation where `isSelectedSiteVirtual` actually contained 0 or 1 instead of false or true... So I thought the truthy equality was "safer" in that sense. But I could look at the code and see if `isSelectedSiteVirtual` takes on `true`/`false`

MatthewMeacham: Took a look, the PHP takes of doing a `(boolean)` cast so we should be good to do `===`

CollinsSpencer: Cool cool. I was just logging it out, and it seemed to always be true or false. Thanks for double-checking.

PR #342, opened 2020-05-29
Appointment Ready Improvements

MatthewMeacham:
A few simple changes:
* Added a "Type" field to the appointment ready email that is sent to administrators. It will be either "residential" or "non-residential" (this can probably eventually be removed, but it handy for testing right now)
* The appointment ready email is sent to a different email depending on if the appointment is a residential or a non-residential appointment
* The uploaded documents are now filed into one of two top-level folders in the BLOB storage: "residential" or "non-residential" to help administrators manage appointments.

### DIFF
--- a/server/appointment/upload-documents/uploadDocuments.php
+++ b/server/appointment/upload-documents/uploadDocuments.php
@@ -121,11 +121,17 @@ function uploadDocument($token, $firstName, $lastName, $emailAddress, $phoneNumb
 		// Validate the client information
 		$clientInformation = validateClientInformation($token, $firstName, $lastName, $emailAddress, $phoneNumber);
 		$appointmentId = $clientInformation['appointmentId'];
+		$appointmentType = $clientInformation['appointmentType'];
 
 		// Upload the user's file to Azure BLOB Storage
 		$containerName = 'ty2019';
 		$fileContent = fopen($uploadedFileTempName, 'r');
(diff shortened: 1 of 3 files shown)


server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -147,27 +153,23 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
-					<b>Appointment ID:</b> $appointmentId <br/>
-					<b>Phone Number:</b> $phoneNumber <br/>
-					<b>Best Time to Call (if new appointment):</b> $bestTimeToCall <br/> 
-					<b>Email (if provided):</b> $emailAddress";
-				EmailUtilities::sendHtmlFormattedEmail($toEmailsString, 'VITA -- Appointment Ready', $readyMessage);
-				fclose($handle);
-			}
+			$emailJsonString = file_get_contents('./notificationEmails.json');
+			$emailsJson = json_decode($emailJsonString, true);
+			$toEmailsString = $appointmentType === 'residential' ? $emailsJson['residential'] : $emailsJson['non-residential'];
```
jcarrigan2, line 162: Not sure if there is a better variable name for this, it's pretty clear once you get to mailUtilities::sendHtmlFormattedEmail() anyway

MatthewMeacham: The best other name I could think of is just `toEmails`. I suppose the `String` piece of the name was a remnant of the old code where it was less clear what was what (i.e. we had an array for all the to emails that were then joined together to form the string)

PR #346, opened 2020-06-03
Appointment Questions Saving Bug Fix

MatthewMeacham:
Remove the country text question selection if the appointment is a residential appointment. 

This was discovered from a small bug where someone's appointment which should've been a residential (and was scheduled for a residential site) was for some reason appearing as a non-resident appointment. This PR addresses that issue by removing the country selection if the appointment is a residential appointment (similar to how we reset other questions when other options change)

### DIFF
--- a/signup/signupController.js
+++ b/signup/signupController.js
@@ -82,7 +82,7 @@ define('signupController', [], function() {
 				else if(country.treatyType === 'india') answerDatabaseId = "12";
 				else if(country.treatyType === 'treaty') answerDatabaseId = "13";
 				else if(country.treatyType === 'non-treaty') answerDatabaseId = "14";
-				
+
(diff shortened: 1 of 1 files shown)


signup/signupController.js
```diff
@@ -158,6 +158,7 @@ define('signupController', [], function() {
 		};
 
 		$scope.residentialAppointment = function() {
+			$scope.questions[6] = null;
```
jcarrigan2, line 161: I'm assuming this has to be accessed by its index?

MatthewMeacham: Yah, at least it's what the other ones do above. Theoretically, I think we could assign an ID to it in HTML and like query that somehow, but I think that is more work than it's worth for this code since the questions don't really ever change

PR #357, opened 2020-07-10
New Queue

MatthewMeacham:
# Note: We will be removing the public queue completely, and replacing the private queue with this queue once this is approved.
# Second Note: I would like some feedback/ideas on the UI and how to make it look better
After talking with Mac, we determined that the current queue wasn't really ideal for what they wanted to do at the larger sites. This PR changes the queue to be a Kanban-style swimlane board where a volunteer can drag an appointment between each of the statuses. Additionally, volunteers can organize the appointments within the swimlane (although this ordering is lost upon refresh).

Volunteers can still click on an appointment to view additional information, have access to the notes, or cancel/incomplete the appointment.

image

# Testing
* Log in
* Navigate to `localhost/queue/new/`
* Play around with it


server/accessors/appointmentAccessor.class.php
```diff
@@ -57,7 +57,7 @@ private function resetFieldsInServicedAppointmentByAppointmentId($appointmentId)
 
 		$query = "UPDATE ServicedAppointment
 			SET timeIn = NULL, timeReturnedPapers = NULL, timeAppointmentStarted = NULL, timeAppointmentEnded = NULL,
-				completed = FALSE, cancelled = FALSE, servicedByStation = NULL
+				completed = NULL, cancelled = FALSE, servicedByStation = NULL
```
jcarrigan2, line 60: What's the reason for this change?

queue/new/queue.php
```diff
@@ -47,15 +47,15 @@ class="dcf-w-100% dcf-input-text"
 		<div ng-if="appointments.length > 0" ng-cloak>
 			<!-- Swimlane headers -->
 			<div class="dcf-grid dcf-grid-fifths@md dcf-col-gap-2">
-				<div class="dcf-d-flex dcf-jc-center"><h5>Awaiting</h5></div>
-				<div class="dcf-d-flex dcf-jc-center"><h5>Checked-In</h5></div>
-				<div class="dcf-d-flex dcf-jc-center"><h5>Paperwork Complete</h5></div>
-				<div class="dcf-d-flex dcf-jc-center"><h5>Preparing</h5></div>
-				<div class="dcf-d-flex dcf-jc-center"><h5>Complete</h5></div>
+				<div class="dcf-d-flex dcf-jc-center"><h5>Awaiting ({{awaitingAppointments.length}})</h5></div>
```
jcarrigan2, line 49: nice add

server/accessors/appointmentAccessor.class.php
MatthewMeacham: Good question--I think it was wrong before. This field starts out `null` in the DB, and we use this field to determine if an appointment is completed or not completed (obviously) where `TRUE` means completed successfully, `FALSE` means marked as incomplete, and `NULL` means isn't in a state where it's complete or incomplete. So I think `NULL` makes more sense here since this method literally just resets the fields to what they were at the start (when we first inserted the ServicedAppointment entry)

queue/new/queue.php
MatthewMeacham: Yah, Mac said he used the queue to determine how many people were in each state of the process (to determine if they could handle a walk-in or a second return for someone, etc.), and I figured this would make it easier than manually counting

PR #366, opened 2021-01-18
Set up document upload for one Azure directory per site

jcarrigan2:
To test this locally:

- Log in to portal.azure.com (you may need to accept an email invite first, subject line "UNL VITA (Tax Credit Campaign) invited you to access applications within their organization"). After logging in, make sure you see "UNL VITA (Tax Credit Campaign)" under your name on the right side.
- Select the unlvitadocuments storage account and find Configuration on the left side. Disable "secure transfer required" to allow for transfer from your localhost on HTTP.
- Also on the left side of the unlvitadocuments Storage Account page, find Access Keys, get the connection string and paste it in config.php. In the string, change https to http.
- I think you may need to make PROD = true in config.php.
- **Make sure to sign up under a name that is a conspicuous test alias like "TestFirst TestLast."**

To test this, I hard-coded siteIds from 1-14 (only 13 exist) to see if it could connect to every site's folder. You can see in ISS's folder that it's able to upload more than one document to a folder.

Container names are currently hard-coded based on siteId. We could refactor this code to include lookupNames for Site table, as has been done for Permission and AppointmentType.


server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -125,9 +125,10 @@ function uploadDocument($token, $firstName, $lastName, $emailAddress, $phoneNumb
 		$clientInformation = validateClientInformation($token, $firstName, $lastName, $emailAddress, $phoneNumber);
 		$appointmentId = $clientInformation['appointmentId'];
 		$appointmentType = $clientInformation['appointmentType'];
-
+		$siteId = $clientInformation['siteId'];
+		
 		// Upload the user's file to Azure BLOB Storage
-		$containerName = 'ty2019';
+		$containerName = getContainerName($siteId);
```
MatthewMeacham, line 131: Oh nice! I forgot there were top-level "containers" in Azure BLOB that we can permission access to. This is brilliant--will avoid us having to provision a new storage account for every site.

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -263,6 +263,39 @@ function doesClientInformationMatch($clientInformation, $firstName, $lastName, $
 	return $firstNameMatches && $lastNameMatches && $emailAddressMatches && $phoneNumberMatches;
 }
 
+// container/blob naming rules here https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata
+function getContainerName($siteId) {
+	$containerName = null;
+	if($siteId == 1) {
+		$containerName = "nebraska-east-union";
+	} else if ($siteId == 2) {
```
MatthewMeacham, line 271: nit: I would probably remove that `$containerName` variable, and just have each if statement return their value, i.e.
```php
if ($siteId == 1) {
    return 'nebraska-east-union';
}
```

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -263,6 +263,39 @@ function doesClientInformationMatch($clientInformation, $firstName, $lastName, $
...
+		$containerName = "victor-e-anderson-library";
+	} else if ($siteId == 3) {
+		$containerName = "jackie-gaughan-multicultural-center";
```
MatthewMeacham, line 274: Also, by convention, you should use single quotes for "constant" strings in PHP (i.e. for any string you're not doing interpolation with).

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -263,6 +263,39 @@ function doesClientInformationMatch($clientInformation, $firstName, $lastName, $
...
+	} else if ($siteId == 10) {
+		$containerName = "southeast-community-college";
+	} else if ($siteId == 11) {
+		$containerName = "nebraska-union";
+	} else if ($siteId == 12) {
+		$containerName = "virtual-vita";
+	} else if ($siteId == 13) {
+		$containerName = "student-athlete-virtual-site";
+	}
+	return $containerName;
```
MatthewMeacham, line 296: While this probably won't happen--I think this last return should be something like:
```php
return "unknown-site-$siteId";
```

Or something, just so we have a safe fallback other than returning `null`.

MatthewMeacham (approved):
Some nitpicks, but this looks good

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: Changed that!

PR #367, opened 2021-01-18
2021 documents

jcarrigan2:


appointment/upload-documents/uploadDocuments.php
```diff
@@ -143,8 +144,8 @@ class="dcf-btn dcf-btn-secondary"
 				<fieldset class="dcf-mt-3">
 					<legend>Appointment Ready</legend>
 					<p>
-						After you mark your appointment as 'ready', a tax preparer will begin preparing your return within 48 hours. 
-						If the tax preparer does not have all the necessary documents, they will not be able to prepare your return. 
+						Thank you for uploading your documents. Your appointment is {{appointmentTimeStr}}.
```
MatthewMeacham, line 147: I'm leery to say "Thank you for uploading your documents" in this box since this box is displayed all the time. I think a description of what "is ready" means is more valuable (i.e. "all your documents have been uploaded").

Also, the "Your documents need to be uploaded by DEADLINE (one week in advance)" text--maybe we should move that out of this box (or duplicate it outside of this box), somewhere a little more noticeable, like at the top of the page or  something? This deadline seems like something pretty important

appointment/upload-documents/uploadDocuments.php
```diff
@@ -167,7 +168,10 @@ class="dcf-btn dcf-btn-primary dcf-mt-1"
 						<ul>
 							<b class="dcf-uppercase">Important Note</b>: If you are typing your information into the fillable form, you must "save as" to your device before uploading it or your information will not save.
 							<li><b>Completed <a href ng-click="downloadForm14446()">Form 14446</a></b>. The <b>bottom portion of page two</b> (Part III: Taxpayer Consents) needs to be completed.</li>
-							<li><b>Completed <a href ng-click="downloadResidentIntakeForm()">Form 13614-C</a></b> (Intake Form). Without this form, Lincoln VITA cannot prepare your return.</li>
+							<li><b>Completed <a href ng-click="downloadResidentIntakeForm()">Form 13614-C</a></b> (Intake Form for residents). Without this form, Lincoln VITA cannot prepare your return.</li>
+							<ul>
+								<li>The Spanish version of the Intake Form for residents can be found here: <b><a href ng-click="downloadResidentIntakeFormSpanish()">Form 13614-C (SP)</a></b></li>
```
MatthewMeacham, line 173: It feels odd to me to add another bullet just to present a version of the document in Spanish. I wonder if something like this might sound better? "Completed Form 13614-C (Intake form for residents). This is available in both an <link> English version</link> and a <link>Spanish version</link>. Without this form, Lincoln VITA cannot prepare your return."

Not sure what the exact verbiage would be--probably should be workshopped with Tracy/Dr. Moody, but this is my gut reaction

MatthewMeacham (commented):
Just some text suggestions, but this looks good.

I did not test locally

appointment/upload-documents/uploadDocuments.php
jcarrigan2: I thought the same thing! Tracy sent me that wording, but I offered the same feedback and suggested we omit the thanks and switch the box name to say "Finished uploading your documents?" Hopefully this will be a bit clearer. I think duplicating that text sounds like a good idea, too.

jcarrigan2:
Meach said this is good to merge.

PR #368, opened 2021-01-19
Quick fix for virtual residential appointments

jcarrigan2:


jcarrigan2:
Meach just walked me through this change, so I'll consider it approved.

PR #374, opened 2021-02-04
Merge email functionality

jcarrigan2:
Now that SCs have Azure access, it also makes sense for them to get the emails letting them know clients have marked their appointments as ready. So instead of emailing either the standard or international vita account, now emails for appointments will be going to their respective site coordinator's email. I believe these are actually just subsidiary inboxes of the vita@unl.edu account. The emails (excluding the two existing emails previously mentioned) are as follows:

- Center for People in Need (cpn-vita@unl.edu)
- F Street Community Center (fstreet-vita@unl.edu)
- Asian Community & Cultural Center (accc-vita@unl.edu)
- Victor E Anderson Library (anderson-vita@unl.edu)
- Loren C Eiseley Library (eiseley-vita@unl.edu)
- Bennett Martin Library (bm-vita@unl.edu)
- Southeast Community College (scc-vita@unl.edu

I HAVE NOT ACTUALLY TESTED THIS CODE. I tried to use [this ](https://tripleroi.com/solvedenabling-sendmail-on-localhost-for-mamp-free-version/) to enable mail on localhost with no luck. Since this is a change on such low-level code, I was hoping that it would be acceptable. What are your guys' thoughts? Any success with testing mail functionality in the past?

### DIFF
--- a/server/appointment/upload-documents/uploadDocuments.php
+++ b/server/appointment/upload-documents/uploadDocuments.php
@@ -172,9 +172,11 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
 
 		// Email volunteers saying it's ready to go
 		if (PROD) {
-			$emailJsonString = file_get_contents('./notificationEmails.json');
-			$emailsJson = json_decode($emailJsonString, true);
-			$toEmailsString = AppointmentTypeUtilities::isResidentialAppointmentType($appointmentType) ? $emailsJson['residential'] : $emailsJson['non-residential'];
+			//$emailJsonString = file_get_contents('./notificationEmails.json');
+			//$emailsJson = json_decode($emailJsonString, true);
+			//$toEmailsString = AppointmentTypeUtilities::isResidentialAppointmentType($appointmentType) ? $emailsJson['residential'] : $emailsJson['non-residential'];
+			$siteId = $clientInformation['siteId'];
+			$toEmailString = getToEmailString($siteId);
 
 			$readyMessage = "A client has marked their appointment as ready: <br/>
 				<b>First Name:</b> $firstName <br/>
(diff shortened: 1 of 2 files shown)


server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -335,6 +337,39 @@ function getContainerName($siteId) {
 	return 'server-contingency-site';
 }
 
+function getToEmailString($siteId) {
```
MatthewMeacham, line 340: Not crucial for this PR, but maybe it's worth adding an `email` column on the `Site` table? Although, maybe it's something that we don't want site coordinators to be able to change easily just in case they mess up, and it goes to a defunct/incorrect email

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -184,7 +186,7 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
 				<b>Best Time to Call (if new appointment):</b> $bestTimeToCall <br/> 
 				<b>Email (if provided):</b> $emailAddress <br/>
 				<b>Type:</b> $appointmentType";
```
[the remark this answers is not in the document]
MatthewMeacham: Actually, similarly, now that we split the Azure BLOB storage based on site--it's maybe worth adding the site title to this too? i.e. there's 4 different siteIds that go to `vita@unl.edu`, it'd be kinda inconvenient to search through each four of those (if you even know which 4 they are)

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -335,6 +337,39 @@ function getContainerName($siteId) {
 	return 'server-contingency-site';
 }
 
+function getToEmailString($siteId) {
+	if($siteId == 1) {
```
MatthewMeacham, line 341: nitpick: You could actually use `===` here instead. It's the same as in JS, where `===` will check both the type and the value (so `"1" === 1` is false, but `"1" == 1` is true).

MatthewMeacham (approved):
LGTM--tbh I typically test the email functionality on the server itself. Typically, I would just include my email in the `toEmailsString` so that I received them all, and could check that it was functional.

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: Changed it! Thanks

server/appointment/upload-documents/uploadDocuments.php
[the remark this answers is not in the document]
jcarrigan2: Added both--I considered site for a later PR but forgot about preferred language. Nice

PR #377, opened 2021-02-12
Just for review--not ready to pull yet.

jcarrigan2:


appointment/upload-documents/uploadDocumentsController.js
```diff
@@ -202,6 +237,14 @@ define('uploadDocumentsController', [], function() {
 			});
 		};
 
+		$scope.validateConsent = (consentId) => {
```
jcarrigan2, line 240: If we choose to keep it this way, is there a better way to do this? (Logically or with different operators)

appointment/upload-documents/uploadDocumentsDataService.js
```diff
@@ -69,6 +69,39 @@ define('uploadDocumentsDataService', [], function($http) {
...
+			storeConsent: function(reviewConsent, virtualConsent, signature) {
+				return $http({
+					url: '/server/appointment/upload-documents/uploadDocuments.php',
+					method: 'POST',
+					data: 'action=storeConsent' +
+						`&reviewConsent=${reviewConsent}` +
+						`&virtualConsent=${virtualConsent}`+
+						`&signature=${signature}`,
+					headers: {
+						'Content-Type': 'application/x-www-form-urlencoded'
```
jcarrigan2, line 82: Do you know what this does?

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -187,7 +192,18 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
 				<b>Preferred Language:</b> $preferredLanguage <br/>
 				<b>Email (if provided):</b> $emailAddress <br/>
 				<b>Type:</b> $appointmentType";
-			EmailUtilities::sendHtmlFormattedEmail($toEmailString, 'VITA -- Appointment Ready', $readyMessage);
+			EmailUtilities::sendHtmlFormattedEmail($toEmailString, 'VITA -- Appointment Ready', $volunteerMessage);
+
+			if(!empty($emailAddress)) {
+				// Email client confirmation
+				$clientMessage = "Congratulations, $firstName $lastName, you have successfully marked your appointment as ready!<br/>
```
jcarrigan2, line 199: You can ignore this for now, but if you have any guidance on SMTP, I would love it.

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +213,83 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+
+	try {
+		$DB_CONN->beginTransaction();
+
+		$virtualAppointmentConsentId = insertConsent($reviewConsent, $virtualConsent, $signature);
+		$DB_CONN->commit();
+
+		$response['success'] = true;
+		$response['virtualAppointmentConsentId'] = $virtualAppointmentConsentId;
+		// $response['message'] = AppointmentConfirmationUtilities::generateAppointmentConfirmation($appointmentId);
```
jcarrigan2, line 230: We don't need this `response['message']`, do we? I think this was copied from signup, but we don't go to a confirmation page for uploading so I don't think this is needed at all.

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +213,83 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+	$consentInsert = 'INSERT INTO Appointment (virtualAppointmentConsentId)
+		VALUES (?)';
+	$consentParams = array($virtualAppointmentConsentId);
+
+	$stmt = $DB_CONN->prepare($consentInsert);
+	if(!$stmt->execute($consentParams)){
+		throw new Exception("There was an issue on the server. Please refresh the page and try again.", MY_EXCEPTION);
+	}
+
+	return $DB_CONN->lastInsertId();
```
jcarrigan2, line 290: Do I need to return this?

server/utilities/appointmentConfirmationUtilities.class.php
```diff
@@ -30,9 +30,8 @@ public static function generateAppointmentConfirmation($appointmentId) {
 				$message .= self::residentialInformation();
 			}
 			$message .= self::miscellaneousInformation();
-			$message .= self::selfServiceAppointmentRescheduleInformation($selfServiceAppointmentRescheduleToken);
 		}
```
jcarrigan2, line 33: So clients are emailed the reschedule link, too.

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +213,83 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+	$response['success'] = false;
+
+	try {
+		$appointmentAccessor = new AppointmentAccessor();
+		$appointmentAccessor->addVirtualAppointmentConsentId($appointmentId, $virtualAppointmentConsentId);
+
+		$response['success'] = true;
+	} catch (Exception $e) {
+		$DB_CONN->rollback();
+		$response['message'] = 'An error occurred while trying to record your consent. Please try again in a few minutes.';
```
jcarrigan2, line 252: Do I need this, either?

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +213,83 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+
+		$response['success'] = true;
+		$response['virtualAppointmentConsentId'] = $virtualAppointmentConsentId;
+		// $response['message'] = AppointmentConfirmationUtilities::generateAppointmentConfirmation($appointmentId);
+	} catch (Exception $e) {
+		$DB_CONN->rollback();
+		$response['message'] = 'An error occurred while trying to record your consent. Please try again in a few minutes.';
+	}
+
+	print json_encode($response);
```
jcarrigan2, line 236: do I need this?

jcarrigan2: Now I'm getting lazy--at this point I'm also taking notes for myself to come back and solidify my understanding of how this works. Don't worry about answering trivial questions.

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -6,6 +6,7 @@
 require_once "$root/vendor/autoload.php";
 require_once "$root/server/utilities/emailUtilities.class.php";
 require_once "$root/server/utilities/appointmentTypeUtilities.class.php";
+require_once "$root/server/accessors/appointmentAccessor.class.php";
```
jcarrigan2, line 9: Is this all it takes to include the accessor? This is all I did and it works but I have no clue how all the pieces of this site are glued together so I wanted to make sure.

appointment/upload-documents/uploadDocuments.php
```diff
@@ -73,21 +73,142 @@ class="submit dcf-btn dcf-btn-primary"
...
+						<div ng-show="consentForm.$submitted || consentForm.signature.$touched">
+							<label class="error" ng-show="consentForm.signature.$error.required">This field is required.</label>
+						</div>
+					</li>
+
+					<p ng-if="consentForm.$invalid && consentForm.$submitted" class="error">Your input is invalid, please correct the errors above and re-submit this form</p>
+					<input type="submit" 
+						value="Submit" 
+						class="submit dcf-btn dcf-btn-primary dcf-mt-4"
+						ng-disabled="!consentData.virtualConsent == 1 || consentData.reviewConsent == null || consentData.signature == null || consentData.signature == ''" />
```
jcarrigan2, line 205: Do I need to check this stuff or should I just let form validation do its work? If none of these are here, it is functional and won't allow submit, but this ensures two-way binding has a say, as well.

appointment/upload-documents/uploadDocuments.php
```diff
@@ -73,21 +73,142 @@ class="submit dcf-btn dcf-btn-primary"
...
+						you do not specify the duration of your consent, your consent is valid for one year from the date
+						of signature. If you believe your tax return information has been disclosed or used improperly in a
+						manner unauthorized by law or without your permission, you may contact the Treasury Inspector General
+						for Tax Administration (TIGTA) by telephone at 1-800-366-4484, or by e-mail at complaints@tigta.treas.gov .
+						While the IRS is responsible for providing oversight requirements to Volunteer Income Tax Assistance (VITA) 
+						and Tax Counseling for the Elderly (TCE) programs, these sites are operated by IRS sponsored partners who
+						manage IRS site operations requirements and volunteer ethical standards. In addition, the locations of these 
+						ites may not be in or on federal Property. </p>
+					<li class="form-radio" id="consentVirtual">
+						<label class="form-required">I am agreeing to use this site's Virtual VITA/TCE Process</label>
```
jcarrigan2, line 171: There isn't anything for me to include in the label for the first yes/no question, so I don't know if I should take this out of the label and just leave it as standard text for consistency or what.

The first question (consenting to participate in random IRS quality assurance review of the site) is optional in that they can say yes or no, but for this one (consenting to virtual preparation) they have to say yes. I have made both "required" on the form, but should I make them `class="form-required"` so that they have the little "required" label? Might make clients think they have to say yes.

I mention it elsewhere, but if I take off the controller validation in the ng-disable for the submit button and rely just on form validation, then if they try to submit without filling out the form I can show the `class="error"` label and say something along the lines of "You must answer this question but can say yes or no."

jcarrigan2 (commented):
This has been the first PR of mine that has to do a lot with design, so I'd like your input on a good number of things. Because clients weren't uploading the 14446 and 13614, we're integrating the 14446 into the site itself, asking for and storing their consent.

One thing I don't see how to comment on is on the signup page--there's a checkmark that says "I consent to virtual preparation." Do you think I should just take that out entirely? Also, keep an eye out for where I might've slipped up or forgotten on error handling!

For adding the table, I used

```USE vita;

CREATE TABLE VirtualAppointmentConsent (
	virtualAppointmentConsentId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
	reviewConsent BOOLEAN NOT NULL,
	virtualConsent BOOLEAN NOT NULL, /* virtual is a keyword */
    signature VARCHAR(355) NOT NULL
);
```

then for adding the foreign key column I used:

```USE vita;

ALTER TABLE appointment

	ADD virtualAppointmentConsentId INTEGER UNSIGNED NULL, /*need to create appointment first, then add consent when they go to upload documents*/
	ADD	FOREIGN KEY(virtualAppointmentConsentId) REFERENCES VirtualAppointmentConsent(virtualAppointmentConsentId);
```

Sorry this is all on a branch named after an email lol, you can mostly disregard the email changes for now since I need to figure out how to test that stuff, it's too stressful just deploying and hoping for the best.

appointment/upload-documents/uploadDocumentsController.js
```diff
@@ -49,6 +48,37 @@ define('uploadDocumentsController', [], function() {
...
+		$scope.storeConsent = function() {
+			const reviewConsent = $scope.consentData.reviewConsent;
+			const virtualConsent = $scope.consentData.virtualConsent;
+			const signature = $scope.consentData.signature;
+
+			UploadDocumentsDataService.storeConsent(reviewConsent, virtualConsent, signature).then((response) => {
+				if (typeof response !== 'undefined' && response && response.success){
+					const virtualAppointmentConsentId = response.virtualAppointmentConsentId;
+					const appointmentId = $scope.appointmentId;
+					$scope.addConsentForeignKeyToAppointment(appointmentId, virtualAppointmentConsentId);
```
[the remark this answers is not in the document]
MatthewMeacham: Yah I think I would reverse the foreign key--remove `FK virtualAppointmentConsentId` from `Appointment`, and instead put `FK appointmentId` in the `VirtualAppointmentConsent` table. That feels more natural to me, and could help avoid the odd "have to make a nested call" issue you mentioned.

appointment/upload-documents/uploadDocumentsController.js
```diff
@@ -49,6 +48,37 @@ define('uploadDocumentsController', [], function() {
...
+					//$scope.consentData.completedConsent = true;
+					//document.body.scrollTop = document.documentElement.scrollTop = 0;
+					//NotificationUtilities.giveNotice('Success', 'You are now ready to upload your documents!');
+				} else {
+					NotificationUtilities.giveNotice('Failure', 'There was an error on the server! Please refresh the page in a few minutes and try again.', false);
+				}
+			});
+		};
+
+		$scope.addConsentForeignKeyToAppointment = function(appointmentId, virtualAppointmentConsentId) {
```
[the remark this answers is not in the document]
MatthewMeacham: Hm, that really shouldn't happen. Assuming you switch the FK order like I mention above, you could always put a unique constraint on the `appointmentId` FK and have the DB just fail to the transaction

appointment/upload-documents/uploadDocumentsController.js
```diff
@@ -96,6 +126,11 @@ define('uploadDocumentsController', [], function() {
 						$scope.isResidentialAppointment = response.residentialAppointment;
 						$scope.appointmentTimeStr = response.appointmentTimeStr;
 						$scope.uploadDeadlineStr = response.uploadDeadlineStr.toUpperCase();
+						$scope.appointmentId = response.appointmentId;
+						
+						// Checking to see if they've already filled out consent
+						const virtualAppointmentConsentId = response.consentId;
+						$scope.validateConsent(virtualAppointmentConsentId);
```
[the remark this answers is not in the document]
MatthewMeacham: Hm, I would probably send a value from the BE that says if they've previously consented or not--that way the BE can actually check the values, and we provide a stronger guarantee to the FE that there's actually been consent

appointment/upload-documents/uploadDocumentsDataService.js
MatthewMeacham: IIRC, it adds a header to the HTTP request that tells the recipient that the values are urlencoded in the URL path.

server/appointment/upload-documents/uploadDocuments.php
MatthewMeacham: Yep that's all it takes--this `require_once` is PHP's mechanism for importing other files / their functions

appointment/upload-documents/uploadDocuments.php
MatthewMeacham: I would check them--might as well have that extra guarantee. Also, I think relying on form validation obscures what we actually expect to happen when you read this HTML--so I like being explicit that for this button to work, this is what we expect

appointment/upload-documents/uploadDocuments.php
MatthewMeacham: I would add required to the actual required toggle. And I think in the interest of transparency, I'd probably add some small text beneath the "IRS" one saying "This is optional, but helps ensure Lincoln VITA is meeting the IRS quality guidelines for tax preparation" or something like that--but maybe pitch this question to Tracy.

appointment/upload-documents/uploadDocumentsController.js
```diff
@@ -49,6 +48,37 @@ define('uploadDocumentsController', [], function() {
 		const ACCEPTABLE_FILE_TYPES = ['application/pdf', 'image/jpeg', 'image/png'];
 		const ACCEPTABLE_FILE_EXTENSIONS = ['.pdf', '.jpeg', '.jpg', '.png'];
 
+		$scope.storeConsent = function() {
+			const reviewConsent = $scope.consentData.reviewConsent;
+			const virtualConsent = $scope.consentData.virtualConsent;
+			const signature = $scope.consentData.signature;
```
MatthewMeacham, line 54: You should probably check these variables here again to make sure they're valid, and if not, maybe raise an error with NotificationUtilities, or set some variable and in the HTML show an invalid form box. It's better to avoid the BE call if we know the inputs are invalid on the FE

appointment/upload-documents/uploadDocumentsController.js
MatthewMeacham: If we did choose to keep it this way, and I don't think we should, then you could just do `$scope.consentData.completeConsent = consentId != null`

appointment/upload-documents/uploadDocumentsDataService.js
```diff
@@ -69,6 +69,39 @@ define('uploadDocumentsDataService', [], function($http) {
 				}, (error) => {
 					return null;
 				});
+			},
+			storeConsent: function(reviewConsent, virtualConsent, signature) {
```
MatthewMeacham, line 73: You should really avoid the `function` keyword and use the arrow functions like we do elsewhere in these files--mostly just for consistency, it's tomato-tomato functionally, but the arrow function is the more modern syntax

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
MatthewMeacham: Hold up--what's the flow here? My expectation would be 1) Go to upload documents page, 2) validate my information, 3) If I haven't consented previously, I'm presented with consent page, 4) Can actually upload documents. 

That's what's happening right? I would still include in the BE response a boolean indicating if they've consented previously or not rather than checking this stuff on the FE / initiating another request to the BE

appointment/upload-documents/uploadDocuments.php
```diff
@@ -73,21 +73,142 @@ class="submit dcf-btn dcf-btn-primary"
...
+							<div class="dcf-btn-group" data-toggle="buttons">
+								<label class="dcf-btn dcf-btn-secondary" 
+									name="consentToVirtual" 
+									ng-model="consentData.virtualConsent" 
+									uib-btn-radio="1" 
+									required>Yes
+								</label>
+								<label class="dcf-btn dcf-btn-secondary" 
+									name="consentToVirtual" 
+									ng-model="consentData.virtualConsent"
```
MatthewMeacham, line 182: What happens if someone says no to this field? Can we display a red error saying "We cannot prepare your tax return without your affirmative consent" or whatever? It seems strange to me that saying "No" here just disables the submit button and doesn't give user any reason why

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -187,7 +192,18 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+			EmailUtilities::sendHtmlFormattedEmail($toEmailString, 'VITA -- Appointment Ready', $volunteerMessage);
+
+			if(!empty($emailAddress)) {
+				// Email client confirmation
+				$clientMessage = "Congratulations, $firstName $lastName, you have successfully marked your appointment as ready!<br/>
+					<b>Site:</b> $siteName".(AppointmentTypeUtilities::isVirtualAppointmentType($appointmentType) ? " (Please do not show up to the site for your virtual appointment)" : "")."<br/>
+					<b>Your Phone Number:</b> $phoneNumber <br/>
+					<b>Your Chosen Best Time to Call:</b> $bestTimeToCall <br/> 
+					<b>Your Preferred Language:</b> $preferredLanguage <br/>
+					<b>Appointment Type:</b> $appointmentType";
```
MatthewMeacham, line 204: You should probably remove this from the client email, I don't see a compelling reason to tell them this information

server/appointment/upload-documents/uploadDocuments.php
MatthewMeacham: No--on the sign up page, the "message" is the appointment confirmation message that's sent to the FE. That shouldn't be relevant here

server/appointment/upload-documents/uploadDocuments.php
MatthewMeacham: Yes you need this--this is the piece that actually sends the response. `json_encode($response)` takes the $response array and encodes it to JSON, and the print is the piece that actually sends it to the FE

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +213,83 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
 	echo json_encode($response);
 }
 
+function storeConsent($reviewConsent, $virtualConsent, $signature) {
+	GLOBAL $DB_CONN;
+
+	$response = array();
+	$response['success'] = false;
+
+	try {
```
MatthewMeacham, line 222: You should 100% check the provided variables before actually storing anything in the DB--it's completely possible for someone to send a request not from your webpage, and we don't want to willy-nilly accept those values--we should validate that the values are what we expect

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +213,83 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+	}
+
+	return $DB_CONN->lastInsertId();
+}
+
+
+function insertConsentForeignKeyToAppointment($appointmentId, $virtualAppointmentConsentId) {
+	GLOBAL $DB_CONN;
+
+	$consentInsert = 'INSERT INTO Appointment (virtualAppointmentConsentId)
```
MatthewMeacham, line 281: You'll remove this when you switch the FK direction--but this should've been an UPDATE statement, not an INSERT

server/appointment/upload-documents/uploadDocuments.php
MatthewMeacham: No you didn't need this as nothing should actually be getting inserted, just updated. I probably would've check if any rows were actually changed and raise an error if not. But nothing probably needed to be returned from this function

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
jcarrigan2: Yes, that is the flow. When the data is validated, getClientInformation now also gets the virtualConsentId from the appointment table. If the ID is null, that means they haven't yet consented, and they are presented with the consent page. If there is an ID there, it accepts that as evidence that they've consented, since they aren't able to submit the consent form unless they've consented (said yes to virtual preparation and added some kind of characters to the signature field).

Do you think I should add a "valid" boolean to the consent table? Then before inputting values into the table, check to make sure the consent to virtual prep is true and the signature has some kind of string in it; if both those conditions are true, make the valid boolean true, otherwise make it false?

As it stands, this would be somewhat redundant because the only way their consent is added to the DB is if the consent fields are valid in the first place. Let me know if I communicated this well enough in my comments before and here, and if you think this is the way to go.

appointment/upload-documents/uploadDocumentsController.js
```diff
@@ -13,16 +13,15 @@ define('uploadDocumentsController', [], function() {
...
 		$scope.isResidentialAppointment = true;
 
-		// Agree to virtual preparation variables
-		// For some reason, you can't bind the checkbox to a primitive boolean, it has to be in an object: https://stackoverflow.com/a/23943930
-		$scope.agreeToVirtualPreparationCheckbox = {
-			checked: false
+		// Consent variables
+		$scope.consentData = {
+			reviewConsent: 0,
+			virtualConsent: 0,
```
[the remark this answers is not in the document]
jcarrigan2: I just read online that for PHP to TINYINT in MySQL you should use integers. https://stackoverflow.com/questions/44539117/php-insert-boolean-values-into-mysql

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: Do you mean just appointment type? I figured it was one more chance to let them know their appointment wasn't a real face-to-face thing

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
jcarrigan2: I agree it shouldn't happen, just wasn't sure if I needed to account for that edge case. I added the unique constraint on the new table, good idea!

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
jcarrigan2: Changed it!

appointment/upload-documents/uploadDocuments.php
```diff
@@ -73,21 +73,142 @@ class="submit dcf-btn dcf-btn-primary"
...
+					<li><b>Performing the interview with the taxpayer(s):</b> The Return Preparer will contact the Client by phone to discuss Form 13614 and review all photo identification, social security cards/ITIN letter and income documents that have been uploaded.</li>
+					<li><b>Preparing the tax return:</b> An IRS tax law certified return preparer will be assigned to the Client by the Site Coordinator. The Site Coordinator will assign a Client to a Return Prepare based upon the certification level of the Return Preparer.</li>
+					<li><b>Performing the quality review:</b> The Site Coordinator will review the Client's return.</li>
+					<li><b>Sharing the completed return:</b> The quality reviewed tax return will be uploaded and sent via secured link to the Client. The Return Preparer will call the Client to review the finished return, making in needed edits to the return, and asking the Client to sign Form 8879.</li>
```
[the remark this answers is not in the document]
jcarrigan2: Didn't even occur to me to proofread this. I just copied and pasted the doc she sent, and the wording from that came from the 14446 itself. Thanks for the diligence, you can just remind me to check for typos and maybe I can save you some time!

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
jcarrigan2: Damn, you are sharp! Good catch. Changed those to 0's during testing. Let me know what you think about the 0/1 over true/false.

appointment/upload-documents/uploadDocumentsController.js
jcarrigan2: That's a good thought. Would that just be repeating the ng-disable logic?

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
jcarrigan2: Initializing them as null prob isn't best practice... Can't start them off as zero because that would be valid input for the first value, and I don't want that answer to be initialized for them (even though the form should make them select something anyway). Set it = -1? Or just try a t/f implementation

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: Thanks for lmk!!

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
MatthewMeacham: I wouldn't add a valid boolean to the consent table--I would sooner pull the values from the table, and check them in PHP, and then include a `consented` boolean in the response from `getClientInformation`. Ik you say "they can't submit form if it's not valid"--but someone could just CURL the endpoint with invalid data and bypass this check the way it stands. I would check the information on the BE to avoid this situation.

server/appointment/upload-documents/uploadDocuments.php
MatthewMeacham: Yah you do want this--this is the error that could be displayed in the `NotificationUtilities.giveNotice` if there was a failure

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
MatthewMeacham: I would use true/false in the FE and then convert it to the 1/0 if necessary on the BE. I think true/false is way easier to work with and more expected--no reason to force the FE to follow some strange convention on the BE

And is there anything wrong with initializing them as null? i.e. that functions, right?

appointment/upload-documents/uploadDocumentsController.js
MatthewMeacham: Yep

server/appointment/upload-documents/uploadDocuments.php
MatthewMeacham: Yah that's all I thought should be deleted.. Ig you give a good point. What were the values that would be displayed for $appointmentType? I thought it was the "lookupType" name which isn't exactly user-friendly, that's what we use for programmatic convenience... If it's the actual name (i.e. the nice-looking one), then I think we're fine including it here

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
jcarrigan2: I see, thanks. Made that change.

appointment/upload-documents/uploadDocumentsController.js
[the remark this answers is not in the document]
jcarrigan2: I converted them back to booleans on the BE now. Initializing as null does function!

appointment/upload-documents/uploadDocuments.php
jcarrigan2: Yeah, good idea

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: No, not very user-friendly, I took it out

appointment/upload-documents/uploadDocuments.php
jcarrigan2: Okay--label on the first question is back on. I think it looks kinda awkward, but it's probably not that big of a deal.

jcarrigan2:
Outstanding points of interest:

- I should probably take out the "I consent to virtual preparation." checkbox on the signup page, right?
- Asking Tracy if she wants to change the wording for the signature input ("TYPE YOUR NAME HERE")
- Making sure my data validation before storing the consent is sufficient (line 220 on the server-side .php file)
- Checking to see if a row is updated after inserting consent--I didn't see this or the data validation in the analogous appointment storing procedure, so I will probably table this for now.
- Double-checking the email wording and proof-reading the wording (I can handle these)

Here's what I used to make the table:
```
USE vita;

CREATE TABLE VirtualAppointmentConsent (
	virtualAppointmentConsentId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
	reviewConsent BOOLEAN NOT NULL,
	virtualConsent BOOLEAN NOT NULL, /* virtual is a keyword */
    signature VARCHAR(355) NOT NULL,
    appointmentId INTEGER UNSIGNED NOT NULL,
    UNIQUE (appointmentId),
    FOREIGN KEY(appointmentId) REFERENCES appointment(appointmentId)
);
```

appointment/upload-documents/uploadDocuments.php
```diff
@@ -73,21 +73,144 @@ class="submit dcf-btn dcf-btn-primary"
...
+						<div ng-show="consentForm.$submitted || consentForm.signature.$touched">
+							<label class="error" ng-show="consentForm.signature.$error.required">This field is required.</label>
+						</div>
+					</li>
+
+					<p ng-if="consentForm.$invalid && consentForm.$submitted" class="error">Your input is invalid, please correct the errors above and re-submit this form</p>
+					<input type="submit" 
+						value="Submit" 
+						class="submit dcf-btn dcf-btn-primary dcf-mt-4" 
+						ng-disabled="!consentData.virtualConsent == true || consentData.reviewConsent == null || consentData.signature == null || consentData.signature == ''" />
```
MatthewMeacham, line 207: This should be `consentData.virtualConsent != true || .... || consentData.signature.trim() == ''`

appointment/upload-documents/uploadDocumentsController.js
```diff
@@ -49,6 +48,27 @@ define('uploadDocumentsController', [], function() {
 		const ACCEPTABLE_FILE_TYPES = ['application/pdf', 'image/jpeg', 'image/png'];
 		const ACCEPTABLE_FILE_EXTENSIONS = ['.pdf', '.jpeg', '.jpg', '.png'];
 
+		$scope.storeConsent = () => {
```
MatthewMeacham, line 51: small nit: I'd consider calling this `submitConsent` instead.. I'm not a huge fan of `store`, but this is mostly personal preference

appointment/upload-documents/uploadDocumentsController.js
```diff
@@ -49,6 +48,27 @@ define('uploadDocumentsController', [], function() {
...
+			const reviewConsent = $scope.consentData.reviewConsent;
+			const virtualConsent = $scope.consentData.virtualConsent;
+			const signature = $scope.consentData.signature;
+			const appointmentId =  $scope.appointmentId;
+
+			if(virtualConsent === true && signature != null) {
```
MatthewMeacham, line 57: I think you should add `&& signature.trim() !== ''`

appointment/upload-documents/uploadDocumentsDataService.js
```diff
@@ -69,6 +69,24 @@ define('uploadDocumentsDataService', [], function($http) {
...
 				});
+			},
+			storeConsent: function(reviewConsent, virtualConsent, signature, appointmentId) {
+				return $http({
+					url: '/server/appointment/upload-documents/uploadDocuments.php',
+					method: 'POST',
+					data: 'action=storeConsent' +
+						`&reviewConsent=${reviewConsent}` +
+						`&virtualConsent=${virtualConsent}`+
+						`&signature=${signature}`+
```
MatthewMeacham, line 80: What happens if someone has a space in their signature? This still works, right? It should, but worth testing

server/accessors/appointmentAccessor.class.php
```diff
@@ -20,6 +20,19 @@ public function cancelAppointment($appointmentId) {
...
+		$query = "UPDATE Appointment
+			SET virtualAppointmentConsentId = ?
+			WHERE appointmentId = ?";
+
+		$stmt = $DB_CONN->prepare($query);
+		$success = $stmt->execute(array($virtualAppointmentConsentId, $appointmentId));
+		if ($success == false) {
+			throw new Exception('Unable to upload consent records.', MY_EXCEPTION);
+		}
+	}
```
MatthewMeacham, line 35: This can be removed

server/appointment/upload-documents/uploadDocuments.php
MatthewMeacham, line 9: This line can be removed

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +211,50 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
 	echo json_encode($response);
 }
 
+function storeConsent($reviewConsent, $virtualConsent, $signature, $appointmentId) {
+	GLOBAL $DB_CONN;
+
+	$response = array();
+	$response['success'] = false;
+	
+	if($virtualConsent === 'true' && $virtualConsent !== null && $signature !== null && $signature !== '' && $appointmentId !== null) {
```
MatthewMeacham, line 220: I would do `$virtualConsent != null && $virtualConsent === 'true' && $signature != null && trim($signature) !== '' && $appointmentId != null`. 

Just using `!=` for the null checks, and trimming the signature

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +211,50 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+		$reviewConsent = $reviewConsent === 'true' ? 1 : 0;
+		$virtualConsent = $virtualConsent === 'true' ? 1 : 0;
+		try {
+			$DB_CONN->beginTransaction();
+
+			insertConsent($reviewConsent, $virtualConsent, $signature, $appointmentId);
+			$DB_CONN->commit();
```
MatthewMeacham, line 227: Since there's actually only one insertion here, you don't need the beginTransaction/commit--you would only do that if you were inserting multiple times and every single one had to complete successfully for the DB to be in a correct state. Given that there's only one insertion here, we're guaranteed to be in a correct state since there's nothing this could rely on

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +211,50 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+
+			$response['success'] = true;
```
MatthewMeacham, line 229: I think you forgot to put the consented response in here? It looks like you expect a `'consented'` in there according to uploadDocumentsController.js line 60.

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +211,50 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+		} catch (Exception $e) {
+			$DB_CONN->rollback();
```
MatthewMeacham, line 231: This rollback isn't necessary either since we don't need the transaction.

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +211,50 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
...
+			$response['message'] = 'An error occurred while trying to record your consent. Please try again in a few minutes.';
+		}
+	} else {
+		$response['message'] = 'Your consent information was not found to be valid. Please try again in a few minutes.';
```
MatthewMeacham, line 235: Maybe update this message to not say "try again in a few"? Maybe just say it was invalid, double check the information and submit again

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -272,17 +326,21 @@ function validateClientInformation($token, $firstName, $lastName, $emailAddress,
...
 
+	// LEFT JOIN consent so that if consent isn't found, it still returns the rest of the info
 	$query = 'SELECT Client.firstName, Client.lastName, Client.emailAddress, Client.phoneNumber, Client.bestTimeToCall, 
 			DATE_FORMAT(AppointmentTime.scheduledTime, "%W, %M %D at %l:%i %p") AS appointmentTimeStr, 
 			DATE_FORMAT(DATE_SUB(AppointmentTime.scheduledTime, INTERVAL 7 DAY), "%W, %M %D") AS uploadDeadlineStr,
 			AppointmentTime.scheduledTime, Appointment.appointmentId, Appointment.language,
-			AppointmentTime.siteId, AppointmentType.lookupName AS appointmentType, Site.title
+			AppointmentTime.siteId, AppointmentType.lookupName AS appointmentType, Site.title,
+			(VirtualAppointmentConsent.virtualConsent = 1 AND NOT (VirtualAppointmentConsent.signature IS NULL OR /*column is set to NOT NULL. Left it to be sure*/
+			VirtualAppointmentConsent.signature = \'\')) as consented
```
MatthewMeacham, line 336: Hm, I think I would sooner pull out the virtual appointment information and have a PHP function to check `consented`. This way, you could use that same function above in `storeConsent` to return `response['consented']`. Then we could also do the `trim` and whatnot

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -197,10 +211,50 @@ function markAppointmentAsReady($token, $firstName, $lastName, $emailAddress, $p
 	echo json_encode($response);
 }
 
+function storeConsent($reviewConsent, $virtualConsent, $signature, $appointmentId) {
```
MatthewMeacham, line 214: Small complication here... As this code is written, anyone could CURL this endpoint with any random `$appointmentId` and we'd put it in the DB. To prevent against attacks like this, every other function in this file requires the token/firstName/lastName/phone/email and re-checks that information before doing the action requested. We need to do that same thing here

MatthewMeacham (commented):
Looking really good! A few more comments to be addressed

You should also put the SQL into `FullDB.sql` and add the `TRUNCATE` in `FullDbTestData.sql` like the other tables

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: Yeah, that's right.

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: I saw elsewhere on the page, null was checked with ===, not ==. How do you know when to use which?

MatthewMeacham: So, I think when we want to check type+value, we should use `===`, and when we just care about value, we should use `==`. Technically, I think it's tomato/tomato in the case of `null`, but I like the idea of `==` better.. In JS for example, it's better to use `==` when checking for `null` or `undefined` b/c it will match either of them (or something like this, can't rightly remember). 

There's actually also a `is_null($val)` function in PHP that could be used instead too, but I think `==` is fine. Doesn't necessarily need a change.

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: That makes sense, I did something similar in the controller

appointment/upload-documents/uploadDocumentsDataService.js
jcarrigan2: Still works!

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: Yeah, you're right about the == checking for null or undefined. Changed this!

jcarrigan2:
With the new test data, I used the following script to get client info and try logging in. Check the emails and I think I'm good!

```
SELECT Appointment.appointmentId, Client.firstName, Client.lastName, Client.phoneNumber, Client.emailAddress, VirtualAppointmentConsent.reviewConsent,
VirtualAppointmentConsent.virtualConsent, VirtualAppointmentConsent.signature, selfserviceappointmentrescheduletoken.token
	FROM vita.appointment
    JOIN vita.Client ON appointment.clientId = Client.clientId
    JOIN virtualAppointmentConsent on appointment.appointmentId = VirtualAppointmentConsent.virtualAppointmentConsentId
    JOIN selfserviceappointmentrescheduletoken on appointment.appointmentId = selfserviceappointmentrescheduletoken.appointmentId
```

jcarrigan2:
Regarding the 14446 consent checkbox on the signup page, I think it should be left there. In the process outlined in the 14446, it says:

>2. Securing Taxpayer Consent Agreement: _Client agrees by checking a box on the scheduling app_, agreeing to participate in the virtual return preparation process. Clients read and electronically sign Form 14446 with a completed and signed Form 13614, Intake, Interview & Quality Review Sheet when uploading documents.

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -189,3 +190,14 @@ CREATE TABLE UserPermission (
 	FOREIGN KEY(permissionId) REFERENCES Permission(permissionId),
 	CONSTRAINT UNIQUE unique_permission (userId, permissionId)
 );
+
+CREATE TABLE VirtualAppointmentConsent (
+	virtualAppointmentConsentId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
+	createdAt DATETIME NOT NULL DEFAULT NOW(),
+	reviewConsent BOOLEAN NOT NULL,
+	virtualConsent BOOLEAN NOT NULL, /* virtual is a keyword */
+    signature VARCHAR(355) NOT NULL,
```
MatthewMeacham, line 199: Is there any reason you made this a `VARCHAR(355)` instead of `VARCHAR(255)` that (I think) we use elsewhere? It doesn't really matter, just curious

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -189,3 +190,14 @@ CREATE TABLE UserPermission (
...
+    appointmentId INTEGER UNSIGNED NOT NULL,
+    UNIQUE (appointmentId),
```
MatthewMeacham, line 201: nit: It's nice to have a defined constraint name so that we can easily delete it in the future if necessary. If we don't give it a name, the DB assigns it some random name that's harder to find
```suggestion
    CONSTRAINT UNIQUE unique_appointment_id (appointmentId),
```

mysql_scripts/table_creation/FullDB.sql
```diff
@@ -189,3 +190,14 @@ CREATE TABLE UserPermission (
...
+    FOREIGN KEY(appointmentId) REFERENCES appointment(appointmentId)
```
MatthewMeacham, line 202: nit:
```suggestion
    FOREIGN KEY(appointmentId) REFERENCES Appointment(appointmentId)
```

mysql_scripts/test_data/FullDBTestData.sql
```diff
@@ -566,6 +567,36 @@ INSERT INTO Answer (possibleAnswerId, appointmentId, questionId)
...
+INSERT INTO VirtualAppointmentConsent (reviewConsent, virtualConsent, signature, appointmentId)
+	VALUES (0, 0, "signature fo\' testing", @appointment_appointment4Id);
+INSERT INTO VirtualAppointmentConsent (reviewConsent, virtualConsent, signature, appointmentId)
+	VALUES (0, 1, "", @appointment_appointment5Id);
+-- end virtualAppointmentConsent
+
+
+
+-- selfserviceappointmentrescheduletoken
+INSERT INTO SelfServiceAppointmentRescheduleToken (token, appointmentId)
```
MatthewMeacham, line 587: Oh nice, thanks for doing this--drove me nuts that I had to sign up for an appointment every time

server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -254,6 +255,9 @@ function insertConsent($reviewConsent, $virtualConsent, $signature, $appointment
 		throw new Exception("There was an issue on the server. Please refresh the page and try again.", MY_EXCEPTION);
 	}
 }
+function validateConsent($virtualConsent, $signature, $appointmentId) {
```
MatthewMeacham, line 258: nit: maybe consider calling this `hasConsentedToVirtualPreparation` or `consentedToVirtualPreparation`, etc. I would expect a `validate` method to just check inputs and throw an error if something is amiss

MatthewMeacham (commented):
Worth noting that you probably need to open another PR after this one exposing whether the client has given their consent to have their appointment reviewed by the IRS--this way if VITA/IRS asks, they could find it on the appointment management page or whatever

Gonna test this now, and then approve if all runs well

appointment/upload-documents/uploadDocuments.php
```diff
@@ -73,21 +73,144 @@ class="submit dcf-btn dcf-btn-primary"
...
+					<p class="dcf-bold">Request to Review your Tax Return for Accuracy:</p>	
+					<p>To ensure you are receiving quality services and an accurately prepared tax return at the volunteer site,
+						IRS employees randomly select free tax preparation sites for review. If errors are identified, the site
+						will make the necessary corrections. IRS does not keep any personal information from your reviewed tax
+						return and this allows them to rate our VITA/TCE return preparation programs for accurately prepared tax
+						returns. If you do not wish to have your return included as part of the review process, it will not affect
+						the services provided to you at this site. If the site preparing this return is selected, do you consent
+						to having your return reviewed for accuracy, by an IRS employee?
+					</p>
+					<li class="form-radio" id="consentReview">
```
MatthewMeacham, line 127: Hm, you're right--this does look kinda strange. Maybe add a label that says "The IRS may quality review my tax return prepared by Lincoln VITA"

image

appointment/upload-documents/uploadDocuments.php
```diff
@@ -73,21 +73,144 @@ class="submit dcf-btn dcf-btn-primary"
...
+						ng-disabled="consentData.virtualConsent != true || consentData.reviewConsent == null || consentData.signature == null || consentData.signature.trim() == ''" />
+				</form>
 			</div>
 
-			<div ng-if="agreeToVirtualPreparationCheckbox.checked">
+			<div ng-if="consentData.completedConsent === true">
 				<p>
-					You may now upload your documents below. 
+					You may now upload your <b>Form 13614-C</b> and other documents below. 
 					<b>Your documents need to be uploaded by {{uploadDeadlineStr}}.</b>
```
MatthewMeacham, line 214: Maybe worth adding the (one week in advance) here too? I had a heart attack when it said "Feb 17" when I had scheduled my appointment for "Feb 24"--thought it was a mistake

image

MatthewMeacham (approved):
This is super slick! LGTM! That "Virtual appointment consent" form looks _really_ good in the HTML, despite being a wall of text. Really well done

I tested CURL'ing to make sure someone couldn't bypass client info checks and consent validity checks. I also tested putting false directly into the DB and receives the error about "if you think you've already submitted, please contact ....". All this worked as expected!

appointment/upload-documents/uploadDocuments.php
```diff
@@ -73,21 +73,142 @@ class="submit dcf-btn dcf-btn-primary"
...
+								</label>
+							</div>
+						</div>
+						<div ng-show="consentForm.$submitted">
+							<label class="error" ng-show="consentForm.consentToVirtual.$error.required">This field is required.</label>
+						</div>
+					</li>
+
+					<li class="dcf-form-group form-textfield">
+						<label class="dcf-label form-label form-required" for="signature">TYPE YOUR NAME HERE</label>
```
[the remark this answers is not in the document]
jcarrigan2: changed it

mysql_scripts/table_creation/FullDB.sql
jcarrigan2: I think I may have copied it from the first VARCHAR variable I saw: `email` in `User`, which is `VARCHAR(355)`. I changed to 255 for consistency.

mysql_scripts/table_creation/FullDB.sql
jcarrigan2: Changed

mysql_scripts/test_data/FullDBTestData.sql
jcarrigan2: haha np

jcarrigan2:
The following needs to be run on production before deploying:```
CREATE TABLE VirtualAppointmentConsent (
	virtualAppointmentConsentId INTEGER UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
	reviewConsent BOOLEAN NOT NULL,
	virtualConsent BOOLEAN NOT NULL, /* virtual is a keyword */
    signature VARCHAR(255) NOT NULL,
    appointmentId INTEGER UNSIGNED NOT NULL,
    CONSTRAINT UNIQUE unique_appointment_id (appointmentId),
    FOREIGN KEY(appointmentId) REFERENCES Appointment(appointmentId)
);
```

PR #379, opened 2021-02-24
Cancelled appointments making queue look overbooked fix

jcarrigan2:
This PR is pretty straightforward--we don't want to include cancelled appointments when calculating the number of appointments already scheduled. However, I do have one question:

After I wrote the query, I poked around to see how other pages got their appointments from the DB, and I found `appointmentDataPickerService.js` calls `/server/api/appointments/getTimes.php?action=getAppointments&...`, but there is only `getAppointmentTimes` in `getTimes.php`. `getAppointmentTimes` still uses the parameters passed by the service call (`year` and `appointmentType`, but how does this work? `getTimes.php` doesn't have the conventional switch statement, so is `getAppointmentTimes` on line 11 just called irrespective of the function included in the data service's `$http.get` string?

Either way, this function actually looks like it has the same cancel check as I included in `sites.php`. The same is true for `getAppointments` in `/server/queue/queue.php`.

Both `queue.php` and `/server/management/appointments/appointments.php` have a `getAppointments` function in its switch statement--my guess is the code for `getAppointmentTimes` in `getTimes.php` was copied over to `getTimes.php`, or the function was renamed.

Unlike `queue.php` and `getTimes.php`, in `appointments.php` the cancelled status of the appointment is not filtered on, which is good--admins should still be able to see cancelled appointments on that page. All that is to say I don't think this change in `sites.php` will have any unintended consequences.

### DIFF
--- a/components/appointmentPicker/appointmentPickerDataService.js
+++ b/components/appointmentPicker/appointmentPickerDataService.js
@@ -3,8 +3,9 @@ define('appointmentPickerDataService', [], function($http) {
 	function appointmentPickerDataService($http) {
 		return {
 			loadAllAppointments: function(year, appointmentType = "residential") {
+				// Is this able to call anything? I only see getAppointmentTimes in getTimes.php
 				return $http.get(`/server/api/appointments/getTimes.php?action=getAppointments&year=${year}&appointmentType=${appointmentType}`).then(function(response){
-					return response.data;
+					return response.data; 
 				},function(error){
 					return null;
 				});
--- a/components/appointmentPicker/appointmentPickerDataService.js
+++ b/components/appointmentPicker/appointmentPickerDataService.js
@@ -3,8 +3,7 @@ define('appointmentPickerDataService', [], function($http) {
 	function appointmentPickerDataService($http) {
 		return {
 			loadAllAppointments: function(year, appointmentType = "residential") {
-				// Is this able to call anything? I only see getAppointmentTimes in getTimes.php
-				return $http.get(`/server/api/appointments/getTimes.php?action=getAppointments&year=${year}&appointmentType=${appointmentType}`).then(function(response){
+				return $http.get(`/server/api/appointments/getTimes.php?year=${year}&appointmentType=${appointmentType}`).then(function(response){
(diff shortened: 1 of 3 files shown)


components/appointmentPicker/appointmentPickerDataService.js
```diff
@@ -3,8 +3,9 @@ define('appointmentPickerDataService', [], function($http) {
 	function appointmentPickerDataService($http) {
 		return {
 			loadAllAppointments: function(year, appointmentType = "residential") {
+				// Is this able to call anything? I only see getAppointmentTimes in getTimes.php
```
MatthewMeacham, line 6: Yah, so the `getTimes.php` file is a little special. It only has one action (unlike other files where there's multiple, so you have to specify it). Since there's only one, we don't do the switch statement and just call the method outright (like you noted, it's at `getTimes.php` line 11).

So to answer your question, this line works. But yah the `action=getAppointments` is likely the result of a copy-paste mistake and could be removed.

server/management/sites/sites.php
```diff
@@ -82,7 +82,9 @@ function getAppointmentTimesForSite($siteId, $year = null) {
 		FROM AppointmentTime
 		JOIN AppointmentType ON AppointmentTime.appointmentTypeId = AppointmentType.appointmentTypeId
 		LEFT JOIN Appointment ON AppointmentTime.appointmentTimeId = Appointment.appointmentTimeId
+        LEFT JOIN ServicedAppointment ON Appointment.appointmentId = ServicedAppointment.appointmentId
 		WHERE siteId = ?
+			AND (cancelled IS NULL OR cancelled = FALSE)
```
MatthewMeacham, line 87: Ohhh I see, the appointment times weren't _actually_ overbooked, they just appeared to be cuz the number e were reporting on the "Sites Management" page was wrong?

MatthewMeacham (approved):
LGTM

server/management/sites/sites.php
jcarrigan2: Yep

MatthewMeacham (approved):
👍

PR #380, opened 2021-03-08
Add international info to queue

jcarrigan2:
This PR:
- adds visa and appointmentType to the queue page (notably is always shown, even if null, unlike others--should I make it uniform, and if so, which do you think would be best?)
- adds info to the front page for the IRS review this week (kind of hides the signup, but that's still available from the nav bar, and this should prob get pushed through for now for the IRS)
- fixes the bug that wouldn't allow clients to sign up for a certain appointmentTime slot if the only appointments that existed for that slot were all cancelled. (More info in the Slack)

### DIFF
--- a/index.php
+++ b/index.php
@@ -128,7 +128,7 @@ return readfile($documentRoot . $path);
 			</div>
 		</div>
 
-		<div class="dcf-bleed unl-bg-lighter-gray">
+		<div class="dcf-bleed unl-bg-light-gray">
 			<div class="dcf-wrapper">
 				<div class="dcf-grid-halves@md dcf-pt-8 dcf-pb-8">
 					<div>
@@ -138,6 +138,40 @@ return readfile($documentRoot . $path);
 						The UNL Tax Credit Campaign prepares tax returns for low-income, working families and students for free.
 						It is sponsored by the University of Nebraska&ndash;Lincoln's Center for Civic Engagement in conjunction
 						with the Lincoln EITC Coalition.
+					</div>
+				</div>
+			</div>
+		</div>
+
+		<div class="dcf-bleed unl-bg-lighter-gray">
+			<div class="dcf-wrapper">
+				<div class="dcf-grid-halves@md dcf-pt-8 dcf-pb-8">
+					<div>
+						<h2>How does Volunteer Income Tax Assistance (VITA) Program Protect My Civil Rights?</h2>
+					</div>
+					<div>
+						In accordance with federal law and the Department of the Treasury – Internal Revenue Service policy,
+						discrimination against taxpayers on the basis of race, color, national origin (including limited
+						English proficiency), disability, sex (in education programs or activities), age or reprisal is
+						prohibited in programs and activities receiving federal financial assistance. Each and every
+						taxpayer has a set of fundamental rights they should be aware of when dealing with the IRS. Explore
+						your rights and our obligations to protect them.
+						<br><br>The Right to Be Informed
+						<br>The Right to Quality Service
+						<br>The Right to Pay No More than the Correct Amount of Tax
+						<br>The Right to Challenge the IRS’s Position and Be Heard
+						<br>The Right to Appeal an IRS Decision in an Independent Forum
+						<br>The Right to Finality
+						<br>The Right to Privacy
+						<br>The Right to Confidentiality
+						<br>The Right to Retain Representation
+						<br>The Right to a Fair and Just Tax System
+						<br><br>Your Civil Rights are Protected:
+						<br><a href="https://www.irs.gov/pub/irs-pdf/p4053esp.pdf">https://www.irs.gov/pub/irs-pdf/p4053esp.pdf</a>
+						<br><a href="https://www.irs.gov/pub/irs-pdf/p4053csp.pdf">https://www.irs.gov/pub/irs-pdf/p4053csp.pdf</a>
+						<br>
+						<br><br>Taxpayer Bill of Rights:
+						<br><a href="https://www.irs.gov/pub/irs-pdf/p5169.pdf">https://www.irs.gov/pub/irs-pdf/p5169.pdf</a>
 					</div>
 				</div>
 			</div>
--- a/queue/queue.php
+++ b/queue/queue.php
@@ -127,7 +127,7 @@
 			<div ng-cloak>
 				<span><b>Visa Type:</b> {{(selectedAppointment.visa || "None").trim()}}</span>
 			</div>
-			<div ng-if="selectedAppointment.appointmentType != null" ng-cloak>
+			<div ng-cloak>
 				<span><b>Appointment Type:</b> {{selectedAppointment.appointmentType}}</span>
 			</div>
 		</div>
--- a/queue/queue.php
+++ b/queue/queue.php
@@ -124,6 +124,12 @@
 			<div ng-if="selectedAppointment.phoneNumber != null" ng-cloak>
 				<span><b>Phone Number:</b> {{selectedAppointment.phoneNumber}}</span>
 			</div>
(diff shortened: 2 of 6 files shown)


index.php
```diff
@@ -143,6 +143,40 @@ function wdnInclude($path)
...
+						<br>The Right to Privacy
+						<br>The Right to Confidentiality
+						<br>The Right to Retain Representation
+						<br>The Right to a Fair and Just Tax System
+						<br><br>Your Civil Rights are Protected:
+						<br><a href="https://www.irs.gov/pub/irs-pdf/p4053esp.pdf">https://www.irs.gov/pub/irs-pdf/p4053esp.pdf</a>
+						<br><a href="https://www.irs.gov/pub/irs-pdf/p4053csp.pdf">https://www.irs.gov/pub/irs-pdf/p4053csp.pdf</a>
+						<br>
+						<br><br>Taxpayer Bill of Rights:
+						<br><a href="https://www.irs.gov/pub/irs-pdf/p5169.pdf">https://www.irs.gov/pub/irs-pdf/p5169.pdf</a>
```
MatthewMeacham, line 174: These might look better as an unordered list `<ul>`, but maybe not idk

queue/queue.php
```diff
@@ -124,6 +124,12 @@ class="dcf-w-100% dcf-input-text"
 			<div ng-if="selectedAppointment.phoneNumber != null" ng-cloak>
 				<span><b>Phone Number:</b> {{selectedAppointment.phoneNumber}}</span>
 			</div>
+			<div ng-cloak>
+				<span><b>Visa Type:</b> {{(selectedAppointment.visa || "None").trim()}}</span>
+			</div>
+			<div ng-if="selectedAppointment.appointmentType != null" ng-cloak>
```
MatthewMeacham, line 130: FWIW I don't think `appointmentType` would every be `null`

server/api/appointments/getTimes.php
```diff
@@ -45,7 +45,8 @@ function getAppointmentTimesFromDatabase($year, $after, $appointmentType) {
 	GLOBAL $DB_CONN;
 
 	$query = 'SELECT apt.appointmentTimeId, apt.siteId, s.title, DATE(scheduledTime) AS scheduledDate,
-		TIME(scheduledTime) AS scheduledTime, COUNT(DISTINCT a.appointmentId) AS numberOfAppointmentsAlreadyScheduled,
+		TIME(scheduledTime) AS scheduledTime,
+		SUM(IF(a.appointmentId IS NOT NULL AND (sa.cancelled IS NULL OR sa.cancelled = FALSE),1,0)) AS numberOfAppointmentsAlreadyScheduled,
```
MatthewMeacham, line 49: Good catch on this

server/queue/queue.php
```diff
@@ -39,24 +39,32 @@ function getAppointments($date, $siteId) {
...
 		$canViewClientInformation = $USER->hasPermission('view_client_information');
 
-		$query = 'SELECT Appointment.appointmentId, TIME_FORMAT(AppointmentTime.scheduledTime, "%l:%i %p") AS scheduledTime, 
+		$query = 'SELECT t1.appointmentId, TIME_FORMAT(AppointmentTime.scheduledTime, "%l:%i %p") AS scheduledTime, 
 					firstName, lastName, timeIn, timeReturnedPapers, timeAppointmentStarted, timeAppointmentEnded, 
 					completed, cancelled, language, Client.clientId, 
 					(DATE_ADD(AppointmentTime.scheduledTime, INTERVAL 30 MINUTE) < NOW() AND timeIn IS NULL) AS noShow,
-					(AppointmentTime.scheduledTime < Appointment.createdAt) AS walkIn ';
+					(AppointmentTime.scheduledTime < t1.createdAt) AS walkIn , appointmentType.name as appointmentType,
+					t2.visa';
```
MatthewMeacham, line 47: I think you need a space after `visa`, so it should be `t2.visa `. We need the space b/c if we don't append the phoneNumber/emailAddress, we go straight to `FROM Appointment ...` and we need a space b/w the `visa` and `FROM`

You can also remove the space after `AS walkIn` above now

server/queue/queue.php
```diff
@@ -39,24 +39,32 @@ function getAppointments($date, $siteId) {
...
+		$query .= 'FROM Appointment t1
+					LEFT JOIN ServicedAppointment ON t1.appointmentId = ServicedAppointment.appointmentId
+					JOIN Client ON t1.clientId = Client.clientId
+					JOIN AppointmentTime ON t1.appointmentTimeId = AppointmentTime.appointmentTimeId
+					JOIN AppointmentType ON AppointmentTime.appointmentTypeId = AppointmentType.appointmentTypeId
+					LEFT JOIN
+					(SELECT Answer.appointmentId, Answer.questionId, PossibleAnswer.text as visa
+					FROM Answer
+					JOIN PossibleAnswer ON Answer.possibleAnswerId = PossibleAnswer.possibleAnswerId
+					WHERE questionId = 3
```
MatthewMeacham, line 60: I'd prefer you use the lookupName that (I think?) exists in the question table. You'd probably need another sub-query here to find the corresponding id. There's probably precedent for this somewhere else in the codebase.

server/queue/queue.php
```diff
@@ -39,24 +39,32 @@ function getAppointments($date, $siteId) {
...
-					LEFT JOIN ServicedAppointment ON Appointment.appointmentId = ServicedAppointment.appointmentId
-					JOIN Client ON Appointment.clientId = Client.clientId
-					JOIN AppointmentTime ON Appointment.appointmentTimeId = AppointmentTime.appointmentTimeId
-				WHERE DATE(AppointmentTime.scheduledTime) = ?
-					AND AppointmentTime.siteId = ?
-					AND (cancelled IS NULL OR cancelled = FALSE)
-					AND (completed IS NULL OR completed != FALSE)
-					AND Appointment.archived = FALSE
-				ORDER BY AppointmentTime.scheduledTime, firstName, lastName';
+		$query .= 'FROM Appointment t1
```
MatthewMeacham, line 51: I think I'd personally leave this as `Appointment` rather than aliasing it, and then call `t2` the `Answers` or something, idk

server/queue/queue.php
```diff
@@ -39,24 +39,32 @@ function getAppointments($date, $siteId) {
 	try {
 		$canViewClientInformation = $USER->hasPermission('view_client_information');
```
MatthewMeacham, line 40: Don't you want this information showing on the appointment management page too? This is in the queue rn which is good, but I figured we'd want it in the appointment management page too since I think that's more often used to view appointment info?

MatthewMeacham (approved):
LGTM. A few nits that could be taken care of in a different PR

server/queue/queue.php
MatthewMeacham: @jcarrigan2 Actually this one probably needs to be changed before this PR can merge

queue/queue.php
jcarrigan2: Yup, removed the ng-if

server/queue/queue.php
jcarrigan2: Changed it

server/queue/queue.php
jcarrigan2: Yeah, totally right. Probably frustrating for you to add the lookupName only for me to query by ID, sorry. I added that.

server/queue/queue.php
jcarrigan2: Did that, called it VisaAnswer

server/queue/queue.php
jcarrigan2: Good point, I will look into adding that

PR #390, opened 2022-01-14
Login fix

jcarrigan2:
Changes include:

- An old fix for a navigation bar bug that didn't show logged-in users the Queue and Management tabs, perhaps due to a browser update. This merge has been waiting on the confirmation from admin I got today that it resolved the problem on their end. Testable simply by logging in and out, and confirming the navigation bar behaves appropriately.
- Script for populating this year's appointment times. Testing instructions for this can be found below.
- Adding a front-facing cancel/reschedule page to make the feature more transparent/accessible.
- Preventing a client from signing up for another appointment if they have an existing future, uncancelled appointment. If they attempt to sign up with the same first and last name as an existing future, uncancelled appointment, they will be redirected to the new cancel/reschedule page. Tested with different names and appointment types. It is not meant to stop a client hellbent on double signing up, since that would be very difficult to do without having clients create authenticated accounts.
- Directed all Azure storage to one folder, and all email confirmations to one account rather than site-specific email addresses. This is due to decreased virtual appointment volume. I tested this with two appointment types, and the results can be seen in our Azure storage account. If you want to try yourself, you can use the instructions on PR#366.
- Allowing clients to choose between virtual or in-person appointments. I have tested this simply by toggling and then signing up.

To test the DB script:

1. I populated the dev DB's sites table using the following:

`INSERT into
	vita.site (
		title, address, phoneNumber, createdBy, lastModifiedBy
	)
values
	('test1', 'address1', '1234567890', 1, 1),
	('test2', 'address2', '1234567890', 1, 1),
	('test3', 'address3', '1234567890', 1, 1),
	('test4', 'address4', '1234567890', 1, 1),
	('test5', 'address5', '1234567890', 1, 1),
	('test6', 'address6', '1234567890', 1, 1),
	('test7', 'address7', '1234567890', 1, 1),
	('test9', 'address8', '1234567890', 1, 1),
	('test9', 'address9', '1234567890', 1, 1),
	('test10', 'address10', '1234567890', 1, 1);`

2. Logged in using `siteadmin@test.test` and password `test`
3. Changed the ID on lines 8 and 516 (in `insertSite()`) of the script to the value 5 (for the above credentials)
4. Cleared the .htaccess files in `/server/scripts` (NOT in `/server`)
5. Changed all the `dataAlreadyInserted` values to false and uncommented the helper functions
6.  Navigated to this script by entering the directory in my browser's bar
7. Looked for the `SUCCESS` message, then confirmed the appointment entries with the Site Management page!


server/storeAppointment.php
```diff
@@ -165,4 +166,35 @@ function getClientInformationForAppointmentId($appointmentId) {
...
+		// uses same cancelled logic from getTimes.php : getAppointmentTimesFromDatabase()
+		$query = 'SELECT
+			SUM(IF(apt.appointmentId IS NOT NULL AND (sa.cancelled IS NULL OR sa.cancelled = FALSE), 1, 0)) AS numberExistingAppointments
+			FROM vita.appointment apt
+			LEFT JOIN client ON apt.clientId = client.clientId
+			LEFT JOIN servicedappointment sa ON apt.appointmentId = sa.appointmentId
+			LEFT JOIN appointmenttime at on apt.appointmentTimeId = at.appointmentTimeId
+			WHERE firstName = ? AND lastName = ?
+			AND scheduledTime >= NOW()
+			GROUP BY firstName, lastName';
```
MatthewMeacham, line 187: What's with the group by?

cancel/index.php
```diff
@@ -0,0 +1,142 @@
...
+	</div>
+	<div class="dcf-main-content dcf-wrapper">
+		<!-- TemplateBeginEditable name="maincontentarea" -->
+		<div class="dcf-bleed">
+			<div class="dcf-wrapper dcf-pb-8">
+				<p>If you have been redirected to this page, it may be because you have attempted to
+					schedule an appointment even though you have a future appointment already scheduled.
+					Unfortunately VITA Lincoln has limited capacity, so we cannot reserve multiple
+					appointments for individuals. To help us serve as much of the Lincoln community
+					as possible, we kindly ask you to cancel or reschedule your currently scheduled appointment.
```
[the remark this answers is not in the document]
MatthewMeacham: Hm I see now that starting with "It looks like you already have an appointment scheduled." doesn't make any sense if they got to this page via the navbar. I do think the first sentence could be cleaned up a bit, but my suggestion doesn't hold

signup/signupController.js
```diff
@@ -106,19 +110,41 @@ define('signupController', [], function() {
...
-				if (typeof response !== 'undefined' && response && response.success){
-					document.body.scrollTop = document.documentElement.scrollTop = 0;
-					$scope.appointmentId = response.appointmentId;
-					$scope.successMessage = $sce.trustAsHtml(response.message);
+			const firstName = $scope.data.firstName || '';
+			const lastName = $scope.data.lastName || '';
+			SignupService.getExistingClientAppointments(firstName, lastName).then((response) => {
+				if (typeof response !== 'undefined' && response && response.success) {
+					// response.numberExistingAppointments === false when query is empty because none exist
+					if(response.numberExistingAppointments === false || response.numberExistingAppointments['numberExistingAppointments'] === '0') {
```
MatthewMeacham, line 118: Should maybe do this `=== false` on the BE and convert it to a zero. Also, does this really return a stringified number? It doesn't just return `0`?

signup/signupController.js
```diff
@@ -106,19 +110,41 @@ define('signupController', [], function() {
...
+						$scope.hasExistingAppointment = false;
+					} else {
+						$scope.hasExistingAppointment = true;
+					}
```
MatthewMeacham, line 122: The above suggestion will simplify this code even more, but it can be simplified as written too:
```suggestion
					// response.numberExistingAppointments === false when query is empty because none exist
					$scope.hasExistingAppointment = response.numberExistingAppointments !== false && response.numberExistingAppointments['numberExistingAppointments'] !== '0';
```

signup/signupController.js
```diff
@@ -106,19 +110,41 @@ define('signupController', [], function() {
...
 
-					// Send the confirmation email
-					if ($scope.data.email != null && $scope.data.email.length > 0) {
-						$scope.emailConfirmation();
+					// Only allow the client to sign up for an appointment if they don't currently have any uncancelled appointments
```
MatthewMeacham, line 124: Are we worried about denying people with the same name? Do we have any evidence that that doesn't happen?

signup/signupController.js
```diff
@@ -5,6 +5,10 @@ define('signupController', [], function() {
 		$scope.sharedProperties = sharedPropertiesService.getSharedProperties();
 		$scope.successMessage = null;
 		$scope.appointmentId = null; // The id of the client's appointment once they successfully sign up
+		$scope.hasExistingAppointment = null;
```
MatthewMeacham, line 8: Does this actually need to be in `$scope`? It seems like it's only used in the sign-up method

signup/signupController.js
```diff
@@ -106,19 +110,41 @@ define('signupController', [], function() {
...
-						$scope.emailConfirmation();
+					// Only allow the client to sign up for an appointment if they don't currently have any uncancelled appointments
+					if(!$scope.hasExistingAppointment) {
+						SignupService.storeAppointments(data).then((response) => {
+							if (typeof response !== 'undefined' && response && response.success){
+								document.body.scrollTop = document.documentElement.scrollTop = 0;
+								$scope.appointmentId = response.appointmentId;
+								$scope.successMessage = $sce.trustAsHtml(response.message);
+		
+								// Send the confirmation email
```
MatthewMeacham, line 132: Should we start requiring email--unsure if we still have clients w/o email addresses? The new "Cancel/Reschedule Appointment" page's instructions don't make much sense unless the person actually had an email

signup/signupController.js
```diff
@@ -106,19 +110,41 @@ define('signupController', [], function() {
...
+						$scope.hasExistingAppointment = false;
+					} else {
+						$scope.hasExistingAppointment = true;
+					}
 
-					// Send the confirmation email
-					if ($scope.data.email != null && $scope.data.email.length > 0) {
-						$scope.emailConfirmation();
+					// Only allow the client to sign up for an appointment if they don't currently have any uncancelled appointments
+					if(!$scope.hasExistingAppointment) {
```
MatthewMeacham, line 125: This code might read cleaner if you invert some of these ifs (just have to make sure you `return` in this inversion), i.e.
```javascript
if ($scope.hasExistingAppointment) {
    NotificationUtilities.giveNotice('Failure', "You may not sign up for an appointment if you already have an existing one.", false);
    window.location.replace("../cancel/index.php");
    return;
}

// Normal sign-up logic
```

signup/signupController.js
```diff
@@ -106,19 +110,41 @@ define('signupController', [], function() {
...
+								// Send the confirmation email
+								if ($scope.data.email != null && $scope.data.email.length > 0) {
+									$scope.emailConfirmation();
+								}
+							} else {
+								NotificationUtilities.giveNotice('Failure', 'There was an error on the server while storing your appointment. Please refresh the page in a few minutes and try again.', false);
+							}
+						});
+					} else {
+						// TODO finExistingAppoitnment has already thrown an error, do we need that one?
```
MatthewMeacham, line 141: Idk what this comment means

volunteer/index.php
```diff
@@ -105,7 +107,19 @@ function wdnInclude($path)
 		<!-- TemplateBeginEditable name="maincontentarea" -->
 		<div class="dcf-bleed">
 			<div class="dcf-wrapper dcf-pb-8">
-				<p>This page isn't quite ready. Check back later for updates!</p>
+			<p>Children living in poverty need your help. The American Rescue Plan of 2021 has increased the dollar amount of
+			 child tax credits and set the income limit to $0. There are 38,000 non-filers in Nebraska which include 9.600
```
MatthewMeacham, line 111: Comma instead of `.`?
```suggestion
			 child tax credits and set the income limit to $0. There are 38,000 non-filers in Nebraska which include 9,600
```

MatthewMeacham (commented):
Some minor comments, but overall LGTM

cancel/index.php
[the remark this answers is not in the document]
jcarrigan2: Right, this page has a couple different hats. I do like your suggestion would look and read. I'll give it a second thought!

signup/signupController.js
jcarrigan2: You are right, I initially had it here but I ended up throwing all into the storeAppointment method to make the async stuff simpler lol. This $scope variable is a relic of my inexperienced attempts at implementing this functionality. I think it makes sense to move that to a variable!

signup/signupController.js
jcarrigan2: This is a great point--I figured less tech-literate clients would be able to email the address, but they might not have an email! And the phone number is buried at the bottom of the page. I'll bring this up to Dr. Moody for her judgement.

cancel/index.php
```diff
@@ -0,0 +1,142 @@
...
+					<br><br>
+					<b>To cancel or reschedule a previously scheduled appointment, you must find your
+					confirmation email titled "VITA Appointment Confirmation".</b> At the bottom of the email,
+					under the bolded title "Rescheduling or Cancelling your Appointment", you will find a link
+					to rechedule or cancel your appointment.<b> There is no other way to reschedule or cancel an appointment.</b>
```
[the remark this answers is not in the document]
jcarrigan2: lol I don't think so, I was just trying to be clear that the site itself didn't provide any cancellation or rescheduling functionality. Too strong? haha

index.php
```diff
@@ -32,8 +34,10 @@ function wdnInclude($path)
 <div class="dcf-d-flex dcf-flex-wrap dcf-ai-center dcf-wrapper dcf-pt-3 dcf-pb-3 dcf-bold dcf-txt-decor-none unl-font-sans unl-cream" style="background-color: #007197;">
 	<span class="dcf-txt-xs">
 		<span class="dcf-d-inline-block dcf-mt-2 dcf-mb-2 dcf-mr-4">
-			In response to COVID-19, all Lincoln VITA tax appointments will be conducted virtually. 
-			You may schedule a virtual appointment under the 'Need Assistance' tab. The virtual site will be taking appointments through Thursday, April 8th.
+			In response to COVID-19 and local health requirements, Lincoln VITA will be holding face-to-face
+			return preparation in 2022. In addition, Lincoln VITA will have their virtual site (remote preparation)
+			and Facilitated Self Assistance (FSA) available. You may schedule a virtual appointment under the
+			'Need Assistance' tab. The scheduling site will open in mid-January.
```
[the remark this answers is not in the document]
jcarrigan2: This looks great, they sent me some text and I swapped it without much thought. Thanks for the suggestion, I sent it to them!

signup/signupController.js
jcarrigan2: Yeah, this is much better. I changed it and I'll test it out

signup/signupController.js
jcarrigan2: Sometimes I get a bit turned around on when to write a new error message vs. when to use the one returned from the service.

jcarrigan2:
I replaced the inverted if and tested client 1 with existing appt, client 2 with a past appt, and client 2 again with their newly scheduled appointment. All worked as expected! And it looks a lot better too, thanks.

PR #392, opened 2022-02-08
FSA and updated documents

jcarrigan2:
- Updated tax forms and some of the embedded tax form wording. Also updated the site schedule.
- Added a page for FSA--I know the signup page is getting cluttered and the UX isn't ideal. I don't want to create new pages liberally, but I thought it was appropriate here because we can't post the FSA link for clients publicly. I thought it best to create a page for it so that the functionality was front-facing and transparent for clients.
- Added crypto to the questionnaire.
- Fixed the data script issue with ISS appointments being remote (oops).

Question: for uploadDocuments.php, do we want to include the date ("2020", now changed to "2021") in the regex check to see if files have changed? It seems like future developers could overlook it and the checks might fail. False negatives here suck for preparers, false positives (rejecting the file even though it's not the original blank .pdf) seems virtually impossible.

Thanks for the PR!


server/appointment/upload-documents/uploadDocuments.php
```diff
@@ -119,22 +119,22 @@ function uploadDocument($token, $firstName, $lastName, $emailAddress, $phoneNumb
 		}
 
 		// Check if file is Fillable Form 14446 and, if so, that it has changed
-		if (preg_match('/(.*)2020_F14446(.*)\.pdf(.*)/i', $uploadedFileName)) {
```
MatthewMeacham, line 122: Haha I see what you mean about this... This is a hard one, since we need to update these documents yearly too (probably?). The most obvious thing to me would be adding a test that verifies `$currentYear_F14446` passes this regex (just as a reminder for people to update it by failing a test), but we don't have any testing framework in this repo 🙈 

I don't think it's crazy to include the date in this regex to prevent an older version from being uploaded 🤷 That seems a lot simpler / easier than having to have some back-and-forth with the client/tax preparer when the preparer notices the file is the wrong one...

We should probably just put together a "yearly checklist" of things that have to change / be done in prep for the new tax season

server/scripts/2022TaxSeasonDataScript.php
```diff
@@ -445,29 +445,29 @@ function insertInternationalStudentScholarSiteData() {
 		$tuesdayDates = getWeeklyDatesFromRange('2022-03-01', '2022-04-05');
 		foreach ($tuesdayDates as $date) {
 
-			// Virtual China
```
MatthewMeacham, line 448: Wasn't this script run already? You'll have to update these manually via the management page if so--but I still like this update for posterity

MatthewMeacham (approved):

server/appointment/upload-documents/uploadDocuments.php
jcarrigan2: That is a good point. I have notes on a lot of stuff, but this is one that I also noted should be added to a yearly list lol. I will keep the list!

self_assist/index.php
```diff
@@ -0,0 +1,140 @@
...
+		<?php wdnInclude("/wdn/templates_5.1/includes/global/head-1.html"); ?>
+	<!--
+		Membership and regular participation in the UNL Web Developer Network is required to use the UNLedu Web Framework. Visit the WDN site at https://wdn.unl.edu/. Register for our mailing list and add your site or server to UNLwebaudit.
+		All framework code is the property of the UNL Web Developer Network. The code seen in a source code view is not, and may not be used as, a template. You may not use this code, a reverse-engineered version of this code, or its associated visual presentation in whole or in part to create a derivative work.
+		This message may not be removed from any pages based on the UNLedu Web Framework.
+
+		$Id: php.fixed.dwt.php | cf0a670a0fd8db9e20a169941c55c838d7c2ba10 | Wed Dec 12 16:54:41 2018 -0600 | Eric Rasmussen	$
+	-->
+	<!-- TemplateBeginEditable name="doctitle" -->
+	<title>Cancel | VITA Lincoln | University of Nebraska&ndash;Lincoln</title>
```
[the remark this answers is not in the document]
jcarrigan2: Nothing gets past you! Thanks

server/scripts/2022TaxSeasonDataScript.php
jcarrigan2: tldr; this is just for posterity!

You're right that this was run already, I actually ran the script again with only ISS to supplement the already scheduled in-person appointments--so if 3/10 appointments had been scheduled, I used the management page to drop it to 3/3, then changed the script on the server to add 7 in-person appointments to restore the total to the original 10. It wasn't ideal, but it was time-sensitive and I thought this was safer (and maybe faster?) than trying to run and test a script during live tax season. Thanks as always for checking--you never know when I might have misunderstood something.

PR #395, opened 2022-04-15
Add form for file extension to signup page

jcarrigan2:


jcarrigan2:
Overriding merge requirement since it's such a simple change and I was able to test it all on my end.

PR #401, opened 2023-01-19
Queue expansion

benwingerter:


benwingerter:
Mark as incomplete or cancel on the queue appointment page doesn't seem to work even if it's on the last step

benwingerter:
Everything else works great

benwingerter:
> Mark as incomplete or cancel on the queue appointment page doesn't seem to work even if it's on the last step

Actually, mark cancel works. I think it's an ng-disabled logic bug. But this button's effect doesn't occur until page reload

jcarrigan2 (commented):
Need to improve cancel/incomplete functionality at some point

jcarrigan2 (approved):

PR #403, opened 2023-01-20
Crypto in signup

jcarrigan2:


jcarrigan2:
Checked the crypto question on signup and I agree with Ben that it is functional. merging

